PR #26, opened 2014-12-12
add function to set visibility of profile fields + translate from wordpress core

hesfrap:
1. I added the variable args in wp-editprofile.php. It's an array with key of all the fields name, which could been show in the plugin and the value true of all the keys. With the hook "wp_user_frontend_activated_fields" you can manipulate the visibility of the fields if you want. Additionally you can hide the sections "Name", "Contact Info", "Description" and/or "User Profile". If you hide the section Name or Contact Info the fields Username, Nickname and E-Mail are only invisible => they're necessary for save the user profile.

//example for hidding the section "Name" and the field "url" of section "Contact Info"
add_filter('wp_user_frontend_activated_fields','set_profile_activated_fields',10,1);

function set_profile_activated_fields($args){
	$args["showSectionName"] = false;
        $args["url"] = false;
	return $args;
}

2. I added the core translation file "wp-content/languages/admin-x_X.mo". The reason => in this plugin are fields from the core. They should translate the same way as the original fields. As example:
<?php _e( 'First Name' ) ?> //get now the translation of adin-x_X.mo. x = language key, X = language key

### DIFF
--- a/wpuf-editprofile.php
+++ b/wpuf-editprofile.php
@@ -70,12 +70,30 @@ class WPUF_Edit_Profile {
         }
 
         $profileuser = get_user_to_edit( $user_id );
-
         if ( isset( $message ) ) {
             echo '<div class="' . $style . '">' . $message . '</div>';
         }
-        ?>
-        <div class="wpuf-profile">
+        
+        $args = array(
+				"showSectionName" => true,
+				"user_login" => true,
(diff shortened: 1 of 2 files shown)


codeshah:
Hello @hesfrap,
Thank you for your help. Please send the pull request to the develop branch. This request is showing conflict.

PR #64, opened 2016-10-27
Autocomplete of Tags doesn't work for authenticated users.

veer-os:
Autocomplete of Tags doesn't work for authenticated users.


tareq1988:
This was fixed internally. Thanks for the pull request though 👍

PR #186, opened 2017-09-24
Fix for front-end product visibility issue

itowhid06:


tareq1988:
This is how it should work:

```$post_id = 202;
$visibility = 'hidden';

$product = wc_get_product( $post_id );
$product->set_catalog_visibility( $visibility );
$product->save();```

PR #216, opened 2017-10-30
Moved "Custom Fields in post" option from global general options to e…

mahbubme:
…ach field settings


tareq1988:
Maybe we shouldn't remove the setting from the settings page. Think of a situation where I don't want to show anything in frontend at all.

PR #248, opened 2017-11-14
Included dokan vendor registration template

mahbubme:


wpuf-functions.php
```diff
@@ -1388,6 +1388,25 @@ function wpuf_get_post_form_templates() {
 }
 
 /**
+ * Get profile form templates
+ *
+ * @since 2.4
+ *
+ * @return array
+ */
+function wpuf_get_profile_form_templates() {
```
tareq1988: Move this function to pro.

PR #279, opened 2017-12-19
Fix free pack buying errors

itowhid06:


includes/class-user-subscription.php
```diff
@@ -462,4 +469,17 @@ public function has_error( $form_settings = null ) {
 
         return false;
     }
+
```
Sk-Shaikat: Add function documentation.

includes/class-user-subscription.php
```diff
@@ -462,4 +469,17 @@ public function has_error( $form_settings = null ) {
...
+    public static function is_free_pack( $pack_id ) {
```
Sk-Shaikat: Add function documentation.

includes/class-user-subscription.php
Sk-Shaikat: Add function documentation.

itowhid06: Done

PR #441, opened 2018-04-30
changed content of subscription expiration mail fix #427

### LINKED ISSUE #427: Edit Subscription -> Expiration Message
01. I edited the **Expiration Message** for testing and typed in the text box: **Expiration Message**
02. I modified the message for something more detailed, cleared cache and all, and still "**Expiration Message**" is showing.
Would be nice if we could customize the Expiration Message at:
Edit Subscription
-- Subscription Options
-- -- Posting Restriction (tab)
-- -- -- Expiration Message

If we could use replacement tags, like the ones for email templates

itowhid06:


class/frontend-form-post.php
```diff
@@ -508,7 +508,9 @@ function submit_post() {
 
                 // if mail active
                 if ( isset( $user_wpuf_subscription_pack['_enable_mail_after_expired'] ) && $user_wpuf_subscription_pack['_enable_mail_after_expired'] == 'on' ) {
-                    $post_expiration_message = $user_wpuf_subscription_pack['_post_expiration_message'];
+                    $wpuf_user  = wpuf_get_user();
+                    $user_subscription = new WPUF_User_Subscription( $wpuf_user );
```
rafsuntaskin: For consistency we should use the pack_id from $user_subscription instead of using the older $user_wpuf_subscription_pack var

class/frontend-form-post.php
```diff
@@ -508,7 +508,9 @@ function submit_post() {
...
+                    $post_expiration_message = $user_subscription->get_subscription_exp_msg( $user_wpuf_subscription_pack['pack_id'] );
```
rafsuntaskin: For consistency we should use the pack_id from $user_subscription instead of using the older $user_wpuf_subscription_pack var

includes/class-user-subscription.php
```diff
@@ -492,4 +492,20 @@ public static function is_free_pack( $pack_id ) {
...
 
+    /**
+     * Get Expiration Message
+     *
+     * @since 2.8.8
+     * @param $pack_id
+     */
+    public function get_subscription_exp_msg( $pack_id ) {
+        $subs      = new WPUF_Subscription();
+        $sub_pack  = $subs->get_subscription( $pack_id );
```
rafsuntaskin: this is a static function

PR #497, opened 2018-06-28
Internationalized cycle period labels, fix #462

### LINKED ISSUE #462: Please make the expiration period values internationalized in the front end.
On the pack admin add/edit page I see the values in en and I think the logic needs this.
backend

But I need a way to translate the front end `cycle_period`.
/wp-user-frontend/class/subscription.php:828

Thank you! 🍻

mahbubme:


class/subscription.php
```diff
@@ -803,6 +803,21 @@ function get_details_meta_value() {
         return $meta;
     }
 
+    /**
+     * Get cycle label
+     *
```
rafsuntaskin: since missing

class/subscription.php
```diff
@@ -803,6 +803,21 @@ function get_details_meta_value() {
...
+     */
+    function get_cycle_label( $cycle_period, $cycle_number ) {
+        $labels = array(
+            'day' => _n( 'Day', 'Days', $cycle_number, 'wp-user-frontend' ),
+            'week' => _n( 'Week', 'Weeks', $cycle_number, 'wp-user-frontend' ),
+            'month' => _n( 'Month', 'Months', $cycle_number, 'wp-user-frontend' ),
+            'year' => _n( 'Year', 'Years', $cycle_number, 'wp-user-frontend' )
```
rafsuntaskin: Formatting no ok

class/subscription.php
rafsuntaskin: return type missing

PR #604, opened 2019-01-15
Apply login_errors filter

jweisman:
Apply [login_errors filter](https://codex.wordpress.org/Plugin_API/Filter_Reference/login_errors) to allow customization of errors during login.

Now adding this filter will allow the user to override the message before it's displayed.
```
add_filter( 'login_errors', 'custom_login_errors');
```

default

wpuf


tareq1988:
@jweisman I don't think putting the filter inside the loop is a good idea.

jweisman:
Hi @tareq1988 - thanks for your response. I understand the concern although I don't believe there will be a lot of messages in the `$this->login_errors` array. 

The WP `login_errors` filter takes a string. But we can create a `wpuf_login_errors` filter and pass in the `$this->login_errors` array. 

What do you think of that approach?

tareq1988:
You can actually [throw an error](https://github.com/weDevsOfficial/wp-user-frontend/blob/develop/includes/free/class-login.php#L375) with the `wpuf_process_login_errors` filter.

jweisman:
That approach requires the filter to implement all of the login logic including calling the relevant authentication methods. The built-in `login_errors` filter keeps the authentication logic where it belongs and just filters the resulting error message(s).

tareq1988:
Oh, so you were talking about filtering and changing the output texts with filter! The `wpuf_process_login_errors` is actually for adding your own validation and throw an error if needed. WPUF's login errors are filterable and as you can see, we are echoing all errors in a loop. As it's just an array, you can't actually filter the specific error and alter the message as it doesn't have a key.

jweisman:
I appreciate your response.

I'll try to explain the use case here a bit more. Currently if I enter an invalid username or email, I receive an error message such as "Invalid username" or "A user could not be found with this email address". Some say that such a message is bad security practice and that it's better to use something like "Invalid username or password". 

That's exactly what the WordPress `login_error` filter is for. You can see the example in the [Codex doc](https://codex.wordpress.org/Plugin_API/Filter_Reference/login_errors). And if I use the built-in WordPress login widget, I can filter the message. 

The WP User Frontend plugin doesn't currently call that filter before displaying the error message. So by adding the filter to each message in the array, I can replace it with the desired text. Something like:
```
function exl_custom_login_errors($error) {
  if (preg_match('/\binvalid\b/i', $error) {
    return "Either your username, email, or password was incorrect.";
  }
  return $error;
}
 add_filter( 'login_errors', 'exl_custom_login_errors');
```

I hope that makes more sense.

jweisman:
Have you considered accepting this pull request? Our security team requires us to override this login error message, so in the meantime we are manually applying this change every time after we upgrade the plugin. Thanks.

tareq1988:
@jweisman it's been a while, sorry for that. I think this approach solves the problem. What do you think?

```php
public function show_errors() {
    $errors = apply_filters( 'wpuf_login_errors', $this->login_errors );

    if ( $errors ) {
        foreach ( $errors as $error ) {
            echo wp_kses_post( '<div class="wpuf-error">' . __( $error, 'wp-user-frontend' ) . '</div>' );
        }
    }
}
```

jweisman:
Hi @tareq1988 - thanks for picking this up again. Your suggestion can work- I can change my filter to `wpuf_login_errors` instead of `login_errors` and loop through the array. I changed `class-login.php` in my branch if you want to merge (or make the change on your side and we can close the PR).

Thanks! This will be helpful as now we can apply updates to the plugin without making our custom change.

PR #744, opened 2019-11-29
Fix: some Column field issue

rokibulislam:


shabnam611:
@rokibulislam  @itowhid06 

please share with me the link to the issue that I created.

PR #849, opened 2020-04-09
Add CSV support for Email field

raulillana:
Add CSV support for Email field using the same validation on each splitted value.

This will let users introduce a Comma Separated Value for the Email field, supporting multiple values.

So, instead of supporting just one: email@example.com
Now will support many: mail1@domain.com, jhon@doe.net, this@awesome.patch

Best regards! ❤️


raulillana:
Bump @tareq1988

raulillana:
Lacks logic over the validation return.

PR #885, opened 2020-06-11
Fix appearing on child categories of multiple parent categories

hasinur1997:


includes/fields/class-field-post-taxonomy.php
```diff
@@ -215,25 +215,44 @@ public function tax_ajax( $post_id = null ) {
...
 
     public function tax_select( $post_id = null ) {
-        $attr = $this->field_settings;
-
+        $attr     = $this->field_settings;
         $selected = $this->terms ? $this->terms[0] : '';
-
         $required = sprintf( 'data-required="%s" data-type="select"', $attr['required'] );
 
+        $parents      = $this->field_settings['exclude'];
```
ediamin, line 222: @hasinur1997 You've already set `$attr     = $this->field_settings;`, so use $attr in the rest of the method whenever you need `$this->field_settings`

includes/fields/class-field-post-taxonomy.php
```diff
@@ -215,25 +215,44 @@ public function tax_ajax( $post_id = null ) {
...
+        $child_ids     = [];
+        $exclude_type = $this->exclude_type;
+
+        if ( !empty( $parents ) ) {
+            foreach ( $parents as $parent ) {
+                $terms = get_terms( 'category', array(
```
ediamin, line 228: @hasinur1997 We have used both array() and [] in our plugin. This is bad and we should fix it. But for now, if you add new code, for a single php file figure out which notation we're using, is it `array()` or `[]`. Then use that notation in everywhere in that file. For this case, we're using `[]` notation in this file, so we should use `[]` instead of `array()`.

includes/fields/class-field-post-taxonomy.php
```diff
@@ -215,25 +215,44 @@ public function tax_ajax( $post_id = null ) {
...
+                    'parent'     => $parent
+                ) );
+
+                foreach ( $terms as $term ) {
+                    array_push( $child_ids, $term->term_id );
+                }
+            }
+        }
+
+        if ( $this->exclude_type == 'child_of' ) {
```
ediamin, line 239: @hasinur1997 use yoda expression and tripple equal, `'child_of' === $this->exclude_type`

ediamin:
@hasinur1997 There is a setting called `Type` for category fields. This PR only works for Type=Select. Other options like Multiselect, checkbox etc should also be worked.

includes/fields/class-field-post-taxonomy.php
hasinur1997: ok

Issue #899, opened 2020-06-17
Colon missing in multiselect, dropdown, and radio button #v.3.3.1


kabirbd89:
The colon sign (:) still missing for drop down list, radio button, and multi select type data field. #886

PR #902, opened 2020-06-18
Add colon after field label

hasinur1997:


hasinur1997:
https://github.com/weDevsOfficial/wp-user-frontend/issues/899

PR #918, opened 2020-07-08
Add multiple comma separated emails in email field & Perform hooks before mail notifications

raulillana:
- Adds multiple comma separated values in email field
- Removes jQuery Validation on all forms submits
- Fixes using custom variables in mail notifications (when populated via hook) after insert or update

Cherry pick last 2 commits.


tareq1988:
We are using the `type=email` for the email attribute field, which is browser default and will be validated by the browser. Removing the default behavior from the default is not something we are looking for. The email input field has a `multiple` attribute according to the [specification](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email). 

I understand about moving the hook before email notification. But other changes won't be accepted.

tareq1988:
Cherry-picked the hooks and merged.

Issue #916, opened 2020-07-01
Multiple File Upload Field will not show at edit page


onionlian:
I use Pro version, when I use `[wpuf_edit]` to edit post that has 2 files, I cant see the old files there

![image](https://user-images.githubusercontent.com/19660904/86258783-c8479880-bbed-11ea-93e9-c295a4a0a086.png)

but when the post only has one file, it works

![image](https://user-images.githubusercontent.com/19660904/86258836-d8f80e80-bbed-11ea-8934-52f83be5f345.png)

so I cant delete old files when the post has two files

thanks

---

WP User Frontend Pro - personal  3.1.0
WP User Frontend 3.3.1
wordpress  5.4.2

onionlian:
here is my field setting

![image](https://user-images.githubusercontent.com/19660904/86306192-315afa80-bc46-11ea-8825-9cbfced4f6ef.png)
![image](https://user-images.githubusercontent.com/19660904/86306219-42a40700-bc46-11ea-9fe7-7c8e23599393.png)

Issue #913, opened 2020-06-23
Can Not Upload Image With Same Filename


Sarkham:
When you try to set the image using the WPUF form, it appears to upload but then displays an image from a previous post that has the same filename. Checking the media library shows the new image was never uploaded.

This doesn’t happen when setting the featured image in the regular WordPress backend editor – it has no problem uploading a file with the same filename as a previously uploaded file, and it correctly sets the featured image on the post.

any user can edit their post and delete the images from the gallery uploaded by other users to other posts, previous versions solve that by adding -1, -2 … to the end of the image url.

This way any user can delete the entire WordPress upload folder.

ediamin:
Fixed via #919

PR #919, opened 2020-07-16
Fix file upload with same file name

hasinur1997:
If multiple files with the same name were saved, the previous file would be there but the next file would not be saved. This is why these problems have been created.

Fixes https://github.com/weDevsOfficial/wp-user-frontend/issues/916, https://github.com/weDevsOfficial/wp-user-frontend/issues/913

### DIFF
--- a/class/upload.php
+++ b/class/upload.php
@@ -85,8 +85,11 @@ class WPUF_Upload {
         $wpuf_file = isset( $_FILES['wpuf_file'] ) ? $_FILES['wpuf_file'] : []; // WPCS: sanitization ok.
 
         $file_extension = pathinfo( $wpuf_file['name'], PATHINFO_EXTENSION );
+        $hash           = wp_hash( time() );
+        $hash           = substr( $hash, 0, 8 );
+
         $upload = [
-            'name'     => wp_hash( time() ) . '.' . $file_extension,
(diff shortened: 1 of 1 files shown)


class/upload.php
```diff
@@ -84,8 +84,9 @@ public function upload_file( $image_only = false ) {
         //$wpuf_file = isset( $_FILES['wpuf_file'] ) ? array_map( 'sanitize_text_field', wp_unslash( $_FILES['wpuf_file'] ) ) : []; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
         $wpuf_file = isset( $_FILES['wpuf_file'] ) ? $_FILES['wpuf_file'] : []; // WPCS: sanitization ok.
 
+        $file_extension = pathinfo( $wpuf_file['name'], PATHINFO_EXTENSION );
         $upload = [
-            'name'     => $wpuf_file['name'],
+            'name'     => wp_hash( time() ) . '.' . $file_extension,
```
ediamin, line 89: Filename should be like myfile-3a401f66.png. Generate hash like this:
```
$hash = wp_hash( time() );
$hash = substr( $hash, 0, 8 );
```
Then use it like this,
```
'name' => $wpuf_file['name'] . '-' . $hash . '.' . $file_extension
```

Issue #844, opened 2020-04-03
Pay per Post doesn't work if subscription pack is activated


shabnam611:
Description: If the user is subscribed into a subscription pack, user is not redirected to the payment page after submitting a post that is set with Pay per Post.

How to Reproduce the Issue: 

1. Please create two post forms, one for subscription forced and one for pay per post method.
2. Then as a user purchase a subscription pack. 
3. Next try to submit the form where user has to pay for submitting each post.

Current Behavior: The post is published like any other post without redirecting the user to the payment page and the payment process is missing. However, if the user cancels the subscription pack and then tries to submit the form associated with Pay per Post, he is redirected for payment which is the correct behavior. 

Expected Behavior: The Pay per Post should redirect the user to the payment page according to the form settings regardless of any subscription pack activated or not. 

Video: https://www.loom.com/share/e9a1e78c305641c2811a949055923a77

PR #924, opened 2020-07-21
Fix payment pay per post

hasinur1997:
If the user is subscribed into a subscription pack, user is not redirected to the payment page after submitting a post that is set with Pay per Post.

Fix https://github.com/weDevsOfficial/wp-user-frontend/issues/844


ediamin:
Good job both of you @hasinur1997 and @shabnam611

PR #925, opened 2020-07-22
Fix role base permission for accessing a post form

hasinur1997:
If enabled is "Enable role base post" and the user role is not selected, the user can still view the post form.

### DIFF
--- a/includes/class-frontend-render-form.php
+++ b/includes/class-frontend-render-form.php
@@ -307,7 +307,7 @@ class WPUF_Frontend_Render_Form {
             return;
         }
 
-        if ( ! wpuf_user_can_post( $this->form_settings['roles'] ) ) {
+        if ( ! wpuf_user_has_roles( $this->form_settings['roles'] ) ) {
             echo wp_kses_post( '<div class="wpuf-message">' . __( 'You do not have sufficient permissions to access this form.', 'wp-user-frontend' ) . '</div>' );
 
             return;
--- a/includes/class-frontend-render-form.php
+++ b/includes/class-frontend-render-form.php
@@ -307,6 +307,12 @@ class WPUF_Frontend_Render_Form {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


wpuf-functions.php
```diff
@@ -3545,3 +3545,29 @@ function wpuf_max_upload_size() {
 function wpuf_validate_boolean( $var ) {
     return filter_var( $var, FILTER_VALIDATE_BOOLEAN );
 }
+
+/**
```
ediamin, line 3549: @hasinur1997 Change the code for `wpuf_user_can_post` function to following and update the files where you use it,
```
/**
 * Check user has certain roles
 * 
 * @since WPUF_SINCE
 *
 * @param  array  $roles   Permitted user roles to submit a post
 * @param  int    $user_id User id will submit post
 * 
 * @return bool 
 */
function wpuf_user_has_roles( $roles, $user_id = 0 ) {
    if ( empty( $roles ) ) {
        return false;
    }

    $user = $user_id ? get_userdata( $user_id ) : wp_get_current_user();

    if ( ! empty( array_intersect( $user->roles, $roles ) ) ) {
        return true;
    }

    return false;
}
```

includes/class-frontend-render-form.php
```diff
@@ -307,6 +307,12 @@ public function render_form( $form_id, $post_id = null, $atts = [], $form = null
             return;
         }
 
+        if ( ! wpuf_user_has_roles( $this->form_settings['roles'] ) ) {
+            echo wp_kses_post( '<div class="wpuf-message">' . __( 'You do not have sufficient permissions to access this form.', 'wp-user-frontend' ) . '</div>' );
```
ediamin, line 311: Remove the kses_post and add this
```php
?>
<div class="wpuf-message"><?php esc_html_e( 'You do not have sufficient permissions to access this form.', 'wp-user-frontend'  ); ?></div>
<?php
```

PR #944, opened 2020-08-07
Add post edit link

hasinur1997:
Add post edit link. This link will redirect to post edit form in frontend.

### DIFF
--- a/includes/class-frontend-form-post.php
+++ b/includes/class-frontend-form-post.php
@@ -26,6 +26,10 @@ class WPUF_Frontend_Form extends WPUF_Frontend_Render_Form {
         // form preview
         add_action( 'wp_ajax_wpuf_form_preview', [ $this, 'preview_form' ] );
         $this->set_wp_post_types();
+
+        // enable post edit link for post authors
+        add_filter( 'user_has_cap', [ $this, 'map_capabilities_for_post_authors' ], 10, 4 );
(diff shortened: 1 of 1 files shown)


includes/class-frontend-form-post.php
```diff
@@ -1017,4 +1021,60 @@ public function send_mail_for_guest( $charging_enabled, $post_id, $form_id, $is_
 
         return $response;
     }
+    /**
```
ediamin, line 1024: Use following for doc block. Always include 1. Summery, 2. `@since` - if new function, 3. `@param`, 4. `@return`. Remove the `$context` param, since we're not using it and update the last param in the hook in __construct, `add_filter( 'get_edit_post_link', [ $this, 'edit_post_link' ], 10, 2 );`
```
/**
 * Enable post edit link for post authors
 * 
 * @since WPUF_SINCE
 * 
 * @param string $url
 * @param int    $post_id
 * 
 * @return string
*/
```

includes/class-frontend-form-post.php
```diff
@@ -1017,4 +1021,60 @@ public function send_mail_for_guest( $charging_enabled, $post_id, $form_id, $is_
...
+     * Modify post edit link
+     */
+    public function edit_post_link( $url, $post_id, $context ) {
+
```
ediamin, line 1028: Do not include any extra space after function name.

PR #947, opened 2020-08-13
Fix rolebase permission for post form

hasinur1997:
If any post doesn't turn on 'Enable role base post', the post form couldn't be rendered in frontend.

### DIFF
--- a/includes/class-frontend-render-form.php
+++ b/includes/class-frontend-render-form.php
@@ -307,7 +307,7 @@ class WPUF_Frontend_Render_Form {
             return;
         }
 
-        if ( ! wpuf_user_has_roles( $this->form_settings['roles'] ) ) {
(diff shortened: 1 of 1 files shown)


includes/class-frontend-render-form.php
```diff
@@ -307,7 +307,7 @@ public function render_form( $form_id, $post_id = null, $atts = [], $form = null
             return;
         }
 
-        if ( ! wpuf_user_has_roles( $this->form_settings['roles'] ) ) {
+        if ( 'true' == $this->form_settings['role_base'] && ! wpuf_user_has_roles( $this->form_settings['roles'] ) ) {
```
ediamin, line 310: @hasinur1997 use `wpuf_validate_boolean` for boolean testing like 
```
wpuf_validate_boolean( $this->form_settings['role_base'] )
```
No matter what, do not use double comparison. Use triple comparison `===` in both PHP and JavaScript.

Issue #946, opened 2020-08-12
Submitting product attribute from frontend


shabnam611:
Description: 

If users insert product attribute value like this $500000 from the frontend, the value doesn't appear in the product page. 

How to Reproduce the Issue:

Create an attribute from WooCommerce > Product > Attribute section.

Take a WPUF product type form. Then take the custom attribute field from the taxonomy field area. 

Make the field a text input field. Now if we submit the product type post, the value doesn't appear as a newly created term in the attribute. 

Video: https://www.loom.com/share/5846a749ed55434c8e97c247b25ded6d

camilolunacom:
@shabnam611 thanks for reporting this on our behalf.

I would like to add that this was working on version 3.1.9 of the plugin. It's when updating to version 3.3.1 that we found the issue.

PR #955, opened 2020-08-26
Custom taxonomy input text

hasinur1997:
If custom taxonomy field type is text and not exist the term, it couldn't save the term.

Fix: https://github.com/weDevsOfficial/wp-user-frontend/issues/946


camilolunacom:
Hi @hasinur1997 do you know if this is going to be included on the next release? I would like it to be merged before I include it on my website.

PR #959, opened 2020-08-29
Feature: Add character restriction

hasinur1997:


assets/js/frontend-form.js
```diff
@@ -833,43 +833,43 @@
 
         editorLimit: {
 
-            bind: function(limit, field, type) {
+            bind: function(limit, limit_type, field, type) {
```
ediamin, line 836: Do no add new param in between old params. Always add new param as a last argument. Change the position of `limit_type`, move it to after `type` param and change your code accordingly.

FaraziF (commented):
This new feature is working as well with the expected scenario.

PR #961, opened 2020-09-01
Fix: redirect payment page after post updating

hasinur1997:
After updating post redirect to payment page.

Fix: https://bitbucket.org/wedevs/wpuf-pro/issues/215/pay-per-post-subscription-is-redirecting

### DIFF
--- a/includes/class-frontend-form-post.php
+++ b/includes/class-frontend-form-post.php
@@ -967,7 +967,11 @@ class WPUF_Frontend_Form extends WPUF_Frontend_Render_Form {
             }
         }
 
-        if ( isset( $this->form_settings['enable_pay_per_post'] ) && 'true' === $this->form_settings['enable_pay_per_post'] ) {
+        if ( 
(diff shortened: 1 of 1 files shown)


includes/class-frontend-form-post.php
```diff
@@ -967,7 +967,11 @@ public function send_mail_for_guest( $charging_enabled, $post_id, $form_id, $is_
             }
         }
 
-        if ( isset( $this->form_settings['enable_pay_per_post'] ) && 'true' === $this->form_settings['enable_pay_per_post'] ) {
+        if ( 
+            isset( $this->form_settings['enable_pay_per_post'] )
+                && 'true' === $this->form_settings['enable_pay_per_post']
```
ediamin, line 972: @hasinur1997 `&&` should be under `isset`, fix the indentation here first. Then check boolean value with `wpuf_validate_boolean` function, use `wpuf_validate_boolean( $this->form_settings['enable_pay_per_post']  )`

PR #969, opened 2020-09-10
Add subject to filter

aqjw:
Otherwise, there is no way we can get the subject of the mail from this filter


aqjw:
@jweisman @tareq1988 @adamhughes86 @samholford  Could you add this PR to next release?

PR #970, opened 2020-09-11
Fix: couldn't delete file or image from user profile update

hasinur1997:
Couldn't delete file or image from user profile form.

Fix: https://bitbucket.org/wedevs/wpuf-pro/issues/219/image-and-file-upload-field-is-not

### DIFF
--- a/class/upload.php
+++ b/class/upload.php
@@ -291,7 +291,7 @@ class WPUF_Upload {
         $attachment = get_post( $attachment_id );
 
         // post author or editor role
-        if ( get_current_user_id() == absint( $attachment->post_author ) || current_user_can( 'delete_private_pages' ) ) {
+        if ( get_current_user_id() == absint( $attachment->post_author ) || current_user_can( 'read' ) ) {
             $deleted = wp_delete_attachment( $attachment_id, true );
 
             if ( $deleted ) {


class/upload.php
```diff
@@ -291,7 +291,7 @@ public function delete_file() {
         $attachment = get_post( $attachment_id );
 
         // post author or editor role
-        if ( get_current_user_id() == absint( $attachment->post_author ) || current_user_can( 'delete_private_pages' ) ) {
+        if ( get_current_user_id() == absint( $attachment->post_author ) || current_user_can( 'read' ) ) {
```
tareq1988, line 294: @hasinur1997 the change you've made, doesn't it mean any logged in user can delete anyone's file? That's a serious security issue.

tareq1988 (changes requested):

class/upload.php
hasinur1997: @tareq1988 Can you please suggest me a better way to solve this problem ?

tareq1988: What’s the actual reason the deletion is failing?

hasinur1997: If an image or file is given at the time of registration, it does not have an author id. When user update their profile and try to delete or update the image or file couldn't find the author id. Thats why it is failing to delete.

tareq1988: Then this is not the solution.

When someone is registering, after registration when we have the user ID, we need to change the `post_author` to the newly registered user ID. That means, after registration, they will be the owner of those attachments. Then they will be able to delete their own attachments normally.

PR #974, opened 2020-09-15
Fix: Change tax rate when select another state

hasinur1997:
When change billing address state, tax rate couldn't change.


FaraziF:
@hasinur1997  The tax rate calculates perfectly when changing the state.

Issue #938, opened 2020-07-28
Inconsistency with the WPUF login page Lost Password functionality


mmarj:
**How to reproduce:**
1. Visit login page [wpuf-login] and choose the [Lost Password](https://prnt.sc/tq32b4) option.
2. You will notice the site/page URL is changing after submission of the email/username but it's not showing any visible message so that the user can at least acknowledged that the email notification is sent to the user's email.
3. The same thing is happening for the [WPUF Login Widget](https://prnt.sc/tq33w8) 

Additional Screencast:
Please [check out this video](https://drive.google.com/file/d/1pco8PuzfvQVXejeUEYfG88PuB14ODxHF/view?usp=sharing) to know more details.

Anticosti:
+1 Same issue here

Anticosti:
Would be nice if we could get "**_soon_**" these two messages:

**Before** (?action=lostpassword):
If you forgot your login information, your username and instructions on how to reset your password can be sent to you by email. Please enter your profile email address and submit the request.

**After clicking "Get New Password"** (?checkemail=confirm):
Instructions on how to reset your password have been sent to your email address, you should receive it shortly.

Thank you!

PR #976, opened 2020-09-16
Fix: Lost password reset email message

hasinur1997:
Don't show any confirmation message after password reset email send.

Fix: #938

### DIFF
--- a/includes/free/class-login.php
+++ b/includes/free/class-login.php
@@ -362,7 +362,7 @@ class WPUF_Simple_Login {
                 case 'lostpassword':
                     $checkemail = isset( $getdata['checkemail'] ) ? sanitize_text_field( $getdata['checkemail'] ) : '';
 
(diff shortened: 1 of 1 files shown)


includes/free/class-login.php
```diff
@@ -360,8 +360,15 @@ public function login_form() {
 
             switch ( $action ) {
                 case 'lostpassword':
+                    $checkemail = isset( $getdata['checkemail'] ) ? sanitize_text_field( $getdata['checkemail'] ) : '';
 
-                    $this->messages[] = __( 'Please enter your username or email address. You will receive a link to create a new password via email.', 'wp-user-frontend' );
+                    if ( $checkemail == 'confirm' ) {
```
ediamin, line 365: @hasinur1997 use `if ( 'confirm' === $checkemail ) {`

FaraziF (commented):
Confirmation message successfully shows after resetting the password.

PR #977, opened 2020-09-19
Update class-frontend-form-post.php role base condition

condless:
In order to make the line:
$post_author = get_current_user_id();
to be executed when the condition:
wpuf_user_has_roles( $this->form_settings['roles'] )
is true (the user has the compatible role to access the form)

### DIFF
--- a/includes/class-frontend-form-post.php
+++ b/includes/class-frontend-form-post.php
@@ -605,10 +605,8 @@ class WPUF_Frontend_Form extends WPUF_Frontend_Render_Form {
             }
 
             // the user must be logged in already
-        } elseif ( isset( $this->form_settings['role_base'] ) && $this->form_settings['role_base'] == 'true' ) {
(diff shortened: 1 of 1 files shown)


tareq1988:
@FaraziF the condition needs to be tested. Seems like it was added in commit https://github.com/weDevsOfficial/wp-user-frontend/commit/7bea6967d780af07bf39c2c8d709e613f07a5449 by @hasinur1997. Consult with him if needed.

FaraziF:
> @FaraziF the condition needs to be tested. Seems like it was added in commit [7bea696](https://github.com/weDevsOfficial/wp-user-frontend/commit/7bea6967d780af07bf39c2c8d709e613f07a5449) by @hasinur1997. Consult with him if needed.

I have tested that all roles condition are working perfectly with expected requirements.

tareq1988:
Thank you @condless for your contribution 👍

PR #986, opened 2020-10-01
Enhance: Update transaction label

hasinur1997:
Change the label from 'all' to 'completed'.


tareq1988:
Why do we need to change the label? Can’t we also add completed besides of all? Is there any reason?

I see that if no status is passed, it still defaults to `all`. We can keep all, but add the completed status as well.

hasinur1997:
Support ream ask me to change the label. We can add completed besides of all, but the problem is that transaction data table consist of two different database table's data. It confused me to add completed tab.

Issue #945, opened 2020-08-09
Image field with ACF is throwing error


shabnam611:
Description: After visiting a post from backend, and then if it's edited from the frontend, the image field is getting a image file that he did not submit. 

Steps to Reproduce the Issue:

1. Create a field group in ACF with an image upload field. 

2. Create a WPUF post form and set an image upload field and a featured image field. Using that form, submit only featured image and keep the image upload field blank. Submit the post. 

3. As admin, visit the post from backend and then just save it without making any changes. 

4. As the user, edit the post from frontend and then you will see the image upload field has a file now. 

5. Do nothing and then just update the post. You will see the image file is gone. 

Video: https://www.loom.com/share/94ba1f4fea2a4df381adbd7e778a4307

Expected result: The image file, when it was not uploaded, should not be there while editing.

rajeevkarora:
Just to add, after step 5, a blank WP User Frontend Form is also created erroneously. If you have 2 ACF image fields, then 2 erroneous forms are created.

rajeevkarora:
When you save from the backend, a null value is added for the image to the postmeta by ACF. I think a reasonable fix would be for WP Frontend Pro to ignore empty/null values for this field type and treat them as if they don't exist.

I have also noticed that WPUF does not add a field_key entry to the post meta, which is required by ACF. This may be a separate issue, or it may be related.

PR #998, opened 2020-10-22
Fix: image field conflict with acf image field

hasinur1997:
If make a compatible image field with acf image field, then update the post from backend keep empty the image field it shows a default image in frontend post edit form.

Fix: #945


tareq1988:
@hasinur1997 when you're working on a file, also fix the PHP CS issues there.

hasinur1997:
I will fix it.

PR #1002, opened 2020-10-27
fix: option label not working when & use. tooltip added. fixed #218

zamandevs:


admin/form-builder/assets/js/components/field-option-data/template.php
```diff
@@ -30,6 +30,7 @@
 
             <div class="label">
                 <?php _e( 'Label', 'wp-user-frontend' ); ?>
+                <help-text placement="left" text="<?php _e( 'Don\'t use & or other special character for option label', 'wp-user-frontend' ); ?>" />
             </div>
```
tareq1988, line 34: instead of using `_e()`, use `esc_attr_e()` when using it in a attribute.

PR #1003, opened 2020-10-28
Fix/tax rate not affecting

zamandevs:


tareq1988:
@zamandevs is the function file has any changes related to this fix? Is it phpcs fixes only?

zamandevs:
@tareq1988 Two commit here. one for fix & another for cs. after i created this PR hasinur bhai informed me he did this before on another PR. phpcs fix need to merge.

tareq1988:
@zamandevs There are misaligned arrays in the functions file because maybe you fixed them with `phpcbf` or `php-cs-fixer`. Those need to be aligned correctly manually.

zamandevs:
@tareq1988 I don't see any alignment issue both on my Phpstorm & also github file editor. i am stuck at this point. besides phpstorm CS is from wordpress. thus i dn't see difference. pls guide me in this regard.

PR #1008, opened 2020-11-03
Complete Register/Login/Activation/Passwort Rest senario

DenisCGN:
Hello,

I love this plugin, but the Register/Login/Activation/Passwort Rest szenario is totally unprofellsional.
I dont want my users to see a WORDPRESS Standard page, for email activation, passwordreset etc.
And the look of the login form is not the same as the register form.

thanks,
Denis


tareq1988:
Hey @DenisCGN, your PR has all the historical commits from the past, also now it contains conflicts. Please send a clean PR and then we can review.

PR #1012, opened 2020-11-05
Fix/url field new window not working

zamandevs:
make_clickabe function remove for making anchor tag due to its limitation. [Here is more detail](https://core.trac.wordpress.org/ticket/32992)

### DIFF
--- a/includes/fields/class-field-url.php
+++ b/includes/fields/class-field-url.php
@@ -140,14 +140,8 @@ class WPUF_Form_Field_URL extends WPUF_Form_Field_Text {
                 <?php if ( ! $hide_label ): ?>
                     <label><?php echo esc_html( $field['label'] ); ?>:</label>
                 <?php endif; ?>
-                <?php
-                    echo wp_kses( make_clickable( esc_url_raw( $data ) ), [
-                        'a' => [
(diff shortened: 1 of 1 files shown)


includes/fields/class-field-url.php
```diff
@@ -136,18 +137,12 @@ public function render_field_data( $data, $field ) {
...
-                <?php
-                    echo wp_kses( make_clickable( esc_url_raw( $data ) ), [
-                        'a' => [
-                            'href' => [],
-                            'rel'  => []
-                        ],
-                    ] );
-                ?>
+                <a href="<?php echo esc_url_raw( $data ); ?>"
+                   target="<?php echo $field['open_window'] === 'new' ? '_blank' : ''; ?>"> <?php echo esc_url_raw( $data ); ?> </a>
```
tareq1988, line 145: Should you check if open_window is set or not empty? I'm not sure if it's a definitive value. Also there are empty spaces inside the tag. When it's a _blank target, WP adds a rel attribute with "noreferrer noopener" value. We should also add that for SEO purpose.

https://www.wpbeginner.com/beginners-guide/what-is-relnoopener-in-wordpress-explained/

PR #1023, opened 2020-11-18
70% width is affecting HTML content field's width

mmarj:
Though there is no extra class or input type selector to select custom_html field, it's inheriting the initial width as 70% from [backend form builder](http://prntscr.com/vlors6) and [frontend preview/post page](http://prntscr.com/vlot1c). However, assigning width: 100%, fit-content, or auto; showing as it should be for the custom_html field.


mmarj:
I just noticed this change fall a [negative impact on the WC form/template](http://prntscr.com/vls3b1). So please review the PR closely.
Thanks.

PR #1025, opened 2020-11-23
fix: column field width fixed #246

zamandevs:
issue create on pro [Here](https://github.com/weDevsOfficial/wpuf-pro/issues/246)


FaraziF:
@zamandevs Column field breaking problems have solved

PR #1028, opened 2020-11-27
trial values unset if used once

zamandevs:


class/subscription.php
```diff
@@ -1195,4 +1197,23 @@ public function insert_free_pack_subscribers( $pack_id, $userdata ) {
...
+    /**
+     * Reset trials data if used once
+     *
+     * @since WPUF_PRO
+     *
+     * @param $sub_meta
+     *
+     * @return mixed
+     */
+    public function reset_trial_if_used_once( $sub_meta ) {
```
hasinur1997, line 1210: @zamandevs you can use the function name as `reset_trial` instead of `reset_trial_if_used_once`

tareq1988: The function name suggested by @hasinur1997 seems more logical.

tareq1988:
@zamandevs tab issue exists on `class/subscription.php` file, please fix those.

PR #1041, opened 2020-12-18
Fix/ajax child of not working

zamandevs:


FaraziF:
when update category without child this the last one then category not selected.

FaraziF:
A [notice](https://prnt.sc/xyjupj) appears when updating the category

PR #1051, opened 2021-01-13
non recurring subscription did not work fixed using daily cron

zamandevs:


class/subscription.php
```diff
@@ -1256,4 +1259,55 @@ public function reset_trial( $sub_meta ) {
...
+    }
+
+    /**
+     * Cancel non recurring subs if expired
+     *
+     * @since WPUF_PRO
+     *
+     * @return void
+     */
+    public function cancel_non_recurrent_subscription() {
```
hasinur1997, line 1283: Seems spelling mistake in the function name. It may be `cancel_non_recurring_subscription` instead of `cancel_non_recurrent_subscription`

PR #1053, opened 2021-01-18
fix: address field not working when used with conditional logic

zamandevs:


wpuf-functions.php
```diff
@@ -1069,13 +1069,16 @@ function wpuf_show_custom_fields( $content ) {
                                 }
```
hasinur1997, line 1069: Need to fix coding indentation.

PR #1078, opened 2021-03-13
Make it so we can upload an avatar

jayyx2:
...by disabling a random grecaptcha that never loads and I didn't ask for.


zamandevs:
When a user uses multiple Recaptcha this happened. it's already fixed perfectly a few releases ago. no issue exists in this regard.

PR #1086, opened 2021-03-31
gmap required doesnt work when multiple maps are on same form

zamandevs:


FaraziF:
The gmap require field is not working. Whenever I select the required field.

https://www.loom.com/share/868bfcaf7cb14198b9ac324e8f214e29

PR #1096, opened 2021-04-27
recursive func added for nested array sanitization

zamandevs:


tareq1988:
@zamandevs Please fix the conflict here.

PR #1110, opened 2021-05-26
settings page search implemented

zamandevs:


FaraziF:
The actual result is not coming while searching. [Searching Result](https://www.loom.com/share/d844589cfd63447aa4777b06490eb939)
We can implement searching like Dokan. [Suggestion Video](https://www.loom.com/share/46e2cbe152284f9aba5b2a37737ee882)

tareq1988:
@zamandevs You've used vanilla JS here, better to use jQuery as it's mostly used in all over the plugin.

PR #1116, opened 2021-06-15
Malfunctioning with the Fallback to pay per post feature fixed #1115

### LINKED ISSUE #1115: Malfunctioning with the Fallback to pay per post feature.
**How to reproduce:**
1. Create a subscription form with the post/CPT restriction (say, user can submit 2 Post with the pack)
2. Setup the post form settings > Payment Settings (i)Enable Payments (ii) Force Pack and (iii) Fallback to pay per post
3. Purchase the subscription pack and submit your first post. (You have remaining post 1 as you set the total number post 2)
4. Now, try to make another post to fulfill your quota. You will see it will redirect you to the Fallback to pay per post page to make payment to submit the post.

**Expected Behaviour:**
→ This should let the users fill up their quota according to the Subscription Pack. Once the quota is filled, they should run the `Fallback to pay per post` function. 

**Site Info:**
WordPress with the latest V-5.7.2
WPUF Free- 3.5.16

**Screencast/video:**
Please watch this video to be more precise: https://d.pr/v/abQa7E.

zamandevs:


tareq1988:
@zamandevs I see you've moved the hook after the email notification function. Moving it will make it impossible to change the email contents via a filter. Because if someone adds other actions on this hook, it'll always fire later after the email has been sent. So I don't understand the needs of this PR.

zamandevs:
@tareq1988 This hook is using for reducing post count alongside updating other meta vars & 'wpuf_add_post_redirect' filter is using for redirection where we redirect on payment page based on post count. so on a single request if the user has one post remaining & we are reducing it before redirection then it will redirect to fallback payment page although user has one post remaining.

 if we want to change email content we should consider a filter before email send. as we've email template it seems we don't need any filter here. 
 
 as I've tested carefully it doesn't create any inconsistency with other parts of the application.

PR #1118, opened 2021-06-16
map field full address taken in line with acf map

zamandevs:
companion [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/325)

### DIFF
--- a/includes/class-frontend-render-form.php
+++ b/includes/class-frontend-render-form.php
@@ -307,10 +307,10 @@ class WPUF_Frontend_Render_Form {
             return;
         }
 
-        if ( 
-                isset( $this->form_settings['role_base'] ) 
-                && wpuf_validate_boolean( $this->form_settings['role_base'] ) 
-                && ! wpuf_user_has_roles( $this->form_settings['roles'] ) 
+        if (
+                isset( $this->form_settings['role_base'] )
+                && wpuf_validate_boolean( $this->form_settings['role_base'] )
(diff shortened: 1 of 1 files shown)


includes/class-frontend-render-form.php
```diff
@@ -793,9 +793,13 @@ public static function prepare_meta_fields( $meta_vars ) {
...
                     $map_field_data = $value_name;
 
-                    if ( !empty( $map_field_data ) ) {
-                        list( $data['address'], $data['lat'], $data['lng'] ) = explode( ' || ', $map_field_data );
-                        $meta_key_value[$value['name']]                      = $data;
+                    if ( ! empty( $map_field_data ) ) {
+                        if (stripos( $map_field_data, '||' ) !== false ){
+                            list( $data['address'], $data['lat'], $data['lng'] ) = explode( ' || ', $map_field_data );
+                            $meta_key_value[$value['name']]  = $data;
+                        }else{
```
tareq1988, line 800: Please fix the space here. Also, do we know it'll be JSON data for sure? Maybe someone doesn't use ACF, what will happen then?

tareq1988 (changes requested):

includes/class-frontend-render-form.php
zamandevs: we are storing 3 keys lat,lng & address. acf is storing more fileds alongside with these. so if we store all of these keys it will consistent with acf & without acf. we are sure this data is json format. besides we are checking here for old format.

PR #1124, opened 2021-06-25
existing account section filter improved for client modification

zamandevs:
user can change account section menu slug i.e 
`add_filter('wpuf_account_sections',function ($sections){
       $sections['post']['label'] = 'List';
       $sections['subscription']['label'] = 'Membership';
        return $sections;
   },999);`

### DIFF
--- a/class/frontend-account.php
+++ b/class/frontend-account.php
@@ -183,6 +183,16 @@ class WPUF_Frontend_Account {
                 );
             }
 
+            add_filter(
+                'wpuf_account_sections', function ( $sections ) {
+                    $sections_new = [];
+                    foreach ( $sections as $key => $section ) {
+                        $sections_new[ $section['slug'] ] = $section;
+                    }
+                    return $sections_new;
+                }, 99
+            );
+
             $sections           = wpuf_get_account_sections();
             $current_section    = [];
 
--- a/class/frontend-account.php
+++ b/class/frontend-account.php
@@ -160,28 +160,6 @@ class WPUF_Frontend_Account {
         if ( is_user_logged_in() ) {
             $default_active_tab = wpuf_get_option( 'account_page_active_tab', 'wpuf_my_account', 'dashboard' );
             $section            = isset( $_REQUEST['section'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['section'] ) ) : $default_active_tab;
-            $allowed_posts      = $this->get_allowed_cpt();
-            if ( '' !== $allowed_posts ) {
-                add_filter(
-                    'wpuf_account_sections', function ( $account_sections ) use ( $allowed_posts ) {
-                        foreach ( $allowed_posts as $post_type ) {
-                            $cpt['slug']  = $post_type;
-                            $cpt['label'] = ucfirst( $post_type ) . 's';
-                            array_splice( $account_sections, 1, 0, [ $cpt ] );
-                        }
-                        return $account_sections;
-                    }
-                );
-            } else {
-                add_filter(
-                    'wpuf_account_sections', function ( $account_sections ) {
-                        $post['slug']  = 'post';
-                        $post['label'] = __( 'Posts', 'wp-user-frontend' );
-                        array_splice( $account_sections, 1, 0, [ $post ] );
-                        return $account_sections;
-                    }
-                );
-            }
 
             add_filter(
                 'wpuf_account_sections', function ( $sections ) {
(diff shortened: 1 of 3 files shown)


tareq1988:
I didn't notice it before, but seems like you're using the filters the wrong way.

Filters always return something, but here you've 3 filters with the same name, anonymous functions, who are returning data, but not being stored in a variable. How are they working exactly?

Filters are for enabling other plugins to extend the data before we can use them. But why are we doing that in our own plugin where we can use them directly. What we could do is:

```php
$sections = wpuf_get_account_sections();

// add post types and other stuffs
// re-arrange the array if needed to display and all logics
// then print the sections
```

I think we need to rewrite this code without using 3 filters.

zamandevs:
We have few issues here, yes we can add new filter before printing section. We have another filter in the dashboard which format is slug as key, but here we have numeric keys. so user need to check slug before changing labels thus i remap section where slug will be key.  the last filter is doing this. 

other two filters is adding CPT menu which user save as settings.  we could add this feature after calling the functions but what i've found this function called on other places which may require this added section. 

we should avoid anonymous function as CB but i thought we don't need to remove it in any situation so avoid named function.

not sure if i am missing something on using filter, is so please guide me on right way.

tareq1988:
We could add the post types directly to the `wpuf_get_account_sections()` function before returning. This way we won't have to add any filter. Then here, 

```php
$default_sections = wpuf_get_account_sections();
$sections = [];

// format the sections
foreach ( $default_sections as $key => $label ) {
    $sections[ $key ] = $label;
}

// now pass it to the template
wpuf_load_template(
    'account.php', [
        'sections' => $sections,
        'current_section' => $current_section,
    ]
);
```

We can rewrite the function like this:

```php
function wpuf_get_account_sections() {
    $default_sections = [
        [
            'slug' => 'dashboard',
            'label' => __( 'Dashboard', 'wp-user-frontend' ),
        ],
        [
            'slug' => 'edit-profile',
            'label' => __( 'Edit Profile', 'wp-user-frontend' ),
        ],
        [
            'slug' => 'subscription',
            'label' => __( 'Subscription', 'wp-user-frontend' ),
        ],
        [
            'slug' => 'billing-address',
            'label' => __( 'Billing Address', 'wp-user-frontend' ),
        ],
    ];

    $type_sections = [];
    $post_types = wpuf_get_option( 'cp_on_acc_page', 'wpuf_my_account', [ 'post' ] );

    if ( is_array( $post_types ) && $post_types ) {
        foreach ( $post_types as $post_type ) {
            $post_type_object = get_post_type_object( $post_type );

            $type_sections[] = [
                'slug' => $post_type,
                'label' => $post_type_object->label
            ];
        }
    }

    $sections = array_merge( $type_sections, $default_sections );

    return apply_filters( 'wpuf_account_sections', $sections );
}
```

We can do it like this, what do you think?

zamandevs:
Post types added on function. we need remap section before calling 'wpuf_get_account_sections' otherwise user won't be able to edit labels.

class/frontend-account.php
```diff
@@ -160,28 +160,16 @@ public function shortcode( $atts ) {
...
-                add_filter(
-                    'wpuf_account_sections', function ( $account_sections ) {
-                        $post['slug']  = 'post';
-                        $post['label'] = __( 'Posts', 'wp-user-frontend' );
-                        array_splice( $account_sections, 1, 0, [ $post ] );
-                        return $account_sections;
+
+            add_filter(
+                'wpuf_account_sections', function ( $sections ) {
+                    $sections_new = [];
```
tareq1988, line 166: @zamandevs why do we still need this one? We can run the loop without using the `add_filter`, then pass the new mapped sections to the template.

As the `wpuf_get_account_sections()` function call is on the top, anyone will be able to change the labels with filter.

zamandevs: @tareq1988 Current section is indexed numerically. we remap this index with section slug thus user can change slug without checking the current slug. if we want to keep sections like this we need a filter before/after calling the section function. I am using the previous filter without introducing a new one before calling the function. or we need to add another filter after the calling function if we want to index section slug as key. 

if we remap user can change `add_filter('wpuf_account_sections',function ($sections){ $sections['post']['label'] = 'List'; $sections['subscription']['label'] = 'Membership'; return $sections; },999);` like this otherwise he must check current slug.

tareq1988: Then can we change our main function `wpuf_get_account_sections()` like this, so we won't have to worry about indexes.

```php
$default_sections = [
    'dashboard'       => __( 'Dashboard', 'wp-user-frontend' ),
    'edit-profile'    => __( 'Edit Profile', 'wp-user-frontend' ),
    'subscription'    => __( 'Subscription', 'wp-user-frontend' ),
    'billing-address' => __( 'Billing Address', 'wp-user-frontend' ),
];
```

We can keep it backward compatible by passing it to the template like this:

```php
$default_sections = wpuf_get_account_sections();
$sections = [];

// format the sections
foreach ( $default_sections as $key => $label ) {
    // check backward compatibility
    if ( is_array( $label ) ) {
        $sections[ $label['slug'] ] = $label['label']
    } else {
        $sections[ $key ] = $label;
    }
}
```

zamandevs: Firstly thought like this, but this function was called other places which might break current implementation thus i've remap index here.

PR #1166, opened 2021-08-24
controller added for various email notification

zamandevs:
Companion [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/342)


FaraziF:
An unexpected character is showing in the shortcode.

PR #1192, opened 2021-10-27
Fixed new constructor error

joebailey26:
This fixes an error in the Javascript where it will not allow you to call `swal` without `new`. This error was preventing any popups from showing. Adding the `new` keyword fixes this error.

### DIFF
--- a/assets/js/wpuf-form-builder-components.js
+++ b/assets/js/wpuf-form-builder-components.js
@@ -104,7 +104,7 @@ Vue.component('builder-stage', {
 
             // check if these are already inserted
             if ( this.isSingleInstance( field.template ) && this.containsField( field.template ) ) {
-                swal({
+                new swal({
                     title: "Oops...",
                     text: "You already have this field in the form"
                 });
@@ -117,7 +117,7 @@ Vue.component('builder-stage', {
         delete_field: function(index) {
             var self = this;
 
-            swal({
+            new swal({
                 text: self.i18n.delete_field_warn_msg,
                 type: 'warning',
                 showCancelButton: true,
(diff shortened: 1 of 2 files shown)


sapayth:
Hello @joebailey26,
Thanks for your contribution. But sadly the problem you mentioned cannot be reproduced. Can you please give us step-by-step instructions on what problem you were facing? Also, it would be better if you try our latest version.

And for contribution, please create a new branch for your changes. You can follow [this contribution guide](https://github.com/weDevsOfficial/dokan/blob/develop/CONTRIBUTING.md). Feel free to share your thoughts.

joebailey26:
Hi Sapayth,

Please find a demonstration of the error here:
https://monosnap.com/file/KVPj6Do05J96IxVuiOWUL3YcsuoY9I

I have plugin version 3.5.26 installed.

sapayth:
Hello @joebailey26,
Thanks for the video. Still, the popup, delete, adding new fields, etc. are working on our end. So it would be better to find out the problem first. Would you please try to reinstall the plugin with the new version?

If the problem still occurs, there could be a conflict with other plugins or themes. Let's do a conflict test to be confirmed.
1. Please deactivate all plugins except the WP User Frontend and WP User Frontend Pro.
2. Use a default theme such as twenty-twenty.
3. Clear your site cache as well as the browser cache. Then try again.

Please give it a try and confirm if this solves the problem. Feel free to notify us of any queries.

joebailey26:
Thanks, after deactivating all plugins I can delete things in the WPUF Form, so must be a plugin conflict somewhere

PR #1194, opened 2021-11-11
Add is_array validation on checkbox at class-frontend-render-form

alexgaal:
Hi,

I ran into an issue on a project of mine which I fixed with this pull request. It simply adds `is_array( $value_name )` to the if-condition, I'd suggest `is_countable()` but you would have to update your required PHP version to 7.3.0.

Hope I helped you with this and hopefully it will be merged soon.

### DIFF
--- a/includes/class-frontend-render-form.php
+++ b/includes/class-frontend-render-form.php
@@ -815,7 +815,7 @@ class WPUF_Frontend_Render_Form {
                     break;
 
                 case 'checkbox':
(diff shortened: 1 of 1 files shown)


alexgaal:
Close due a better solution is found

alexgaal:
wrongly closed.

zamandevs:
@alexgaal Thanks for your contribution. we have handled nested values and empty value issues regarding the checkbox. will be on coming release.

PR #1198, opened 2021-11-17
address field inconsistency fixed

zamandevs:


tareq1988:
The name or the functionality of the plugin seems wrong. If necessary, the function name should be `wpuf_get_address( $user_id, $field_name = '' )`. It should be responsible for returning the address, maybe map the deprecated keys also. 

`wpuf_billing_address_remap` is a bad name, and we shouldn't call it.

wpuf-functions.php
```diff
@@ -4055,6 +4013,23 @@ function wpuf_payment_success_page( $data ){
     return $redirect_page;
 }
 
+/**
+ *  Inconsistency with keys, remap keys, Back compat with keys
+ *
+ * @param $address_fields
+ * @return array
+ */
+function wpuf_get_address( $address_fields ) {
```
tareq1988, line 4022: The function name suggests it gets the address from somewhere, like a user or from database. But in reality, it maps the field and returns. It could be something like `wpuf_map_address_fields`.

zamandevs: I've named it wpuf_billing_address_remap and after your suggestion, it renames it to wpuf_get_address. will rename as you suggest.

PR #1201, opened 2021-11-18
Fix/conditional logic remove from selected fields

zamandevs:
[HERE](https://github.com/weDevsOfficial/wpuf-pro/issues/368)

### DIFF
--- a/wpuf-functions.php
+++ b/wpuf-functions.php
@@ -4189,3 +4189,34 @@ function wpuf_pagination( $total_items, $per_page ) {
         return '<div class="wpuf-pagination">' . $page_links . '</div>';
     }
 }
+
+/**
+ * Remove conditional from form builder for selected fields
+ *
+ * @param $settings
+ *
+ * @return array
+ */
+function wpuf_unset_conditional( $settings ) {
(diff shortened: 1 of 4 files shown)


FaraziF:
**Working perfectly as expected:** `we remove the conditional logic from the step field of a post form, we may also remove the conditional logic from the step field of the registration form.`

wpuf-functions.php
```diff
@@ -4012,3 +4012,34 @@ function wpuf_payment_success_page( $data ){
...
+    $remove_cond_field = [ 'action_hook', 'step_start' ];
+
+    $field_settings = array_map(
+        function ( $field ) use ( $remove_cond_field ) {
+            if ( in_array( $field['template'], $remove_cond_field, true ) ) {
+                $index = array_filter(
+                    $field['settings'], function ( $settings ) {
+                    return $settings['name'] === 'wpuf_cond';
+                }
+                );
```
tareq1988, line 4033: Fix the indentation, and also resolve conflicts.

zamandevs: Some PR has conflicts that will resolve after the previous one merges. otherwise, if we resolve this it will conflict again after merging the previous one.

PR #1202, opened 2021-11-18
Fix/template override from child theme

sapayth:
Overriding the dashboard/list.php file from the child theme problem is fixed now. The overridden output should be shown where `[wpuf_dashboard]` shortcode is rendered.

### DIFF
--- a/templates/dashboard.php
+++ b/templates/dashboard.php
@@ -35,7 +35,7 @@
     $meta_name  = [];
     $meta_id    = [];
     $meta_key   = [];
-    $post_data  = [];
+    $template_args  = [];
 
     if ( !empty( $meta ) ) {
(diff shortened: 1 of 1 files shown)


templates/dashboard.php
```diff
@@ -35,6 +35,7 @@
     $meta_name  = [];
     $meta_id    = [];
     $meta_key   = [];
+    $post_data  = [];
```
tareq1988, line 38: There is already a variable inside the loop with nearly the same name `$postdata = get_object_vars( $post );`. Rather it should be like `$template_args` as we are passing it to the template.

tareq1988: @sapayth Please fix this issue.

sapayth: Done

PR #1208, opened 2021-11-29
read only option for custom field enhanced

zamandevs:


FaraziF:
Enhancement request: The required fields will be disabled while the read-only field is enabled.
Screenshot 2021-12-20 at 3 57 18 PM

assets/js/frontend-form.js
```diff
@@ -106,6 +107,14 @@
             });
         },
 
+        handleReadOnly: function(){
+            $('.wpuf-form > .read-only').each( function () {
+                $('input, select, textarea', this).each( function () {
+                     $(this).attr( 'disabled', true );
```
tareq1988, line 113: I haven't run the code, but we are disabling the fields here. Is there any code to enable the disabled ones again? If it works, no problem.

zamandevs: There isn't any code to enable this. as the requirement it is not like a toggle. the field will be disabled if the admin wants to.

DenisCGN:
I like to have that option too, especially to show only the USER_LOGIN without to be able to change it.

PR #1213, opened 2021-12-09
fix: Post form showing blank

### LINKED ISSUE #1218: Forms editing or creation display
Im usinng wordpres 5.8.2 with 8.0 php, the forms are not showing for editing and also can`t create or duplicate any.

blank

sapayth:
We are using a dynamic hook depending on the menu title. If the title got translated, the hook doesn't work for some languages(Español, etc.) and shows blank post form.
Fix #1218 
Similar [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/387) in pro.


tareq1988 (changes requested):
Sorry, but not translating it is also not a solution. It should be translatable, and the hook name will be dynamic too. So you need to bind the hook as a dynamic one.

zamandevs:
If translation need to stay need to handle the label like that 
`$label = strtolower(preg_replace('/\s/', '-$1', __( 'User Frontend', 'wp-user-frontend' )));
  add_action( "load-${label}_page_wpuf-post-forms", [ $this, 'post_forms_builder_init' ] );` 
etc. please make sure both in free and pro also works.

PR #1215, opened 2021-12-13
activated email template enhanced

zamandevs:
Companion [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/384)

### DIFF
--- a/includes/free/class-login.php
+++ b/includes/free/class-login.php
@@ -1091,21 +1091,22 @@ class WPUF_Simple_Login {
         $to         = $user_email;
         $login_url  = wp_login_url();
 
-        /* translators: %s: blogname */
-        $subject = sprintf( __( '[%s] Account has been activated', 'wp-user-frontend' ), $blogname );
-        /* translators: %s: username */
-        $message  = sprintf( __( 'Hi %s,', 'wp-user-frontend' ), $user->user_login ) . "\r\n\r\n";
-        $message .= __( 'Congrats! Your account has been activated. To login visit the following url:', 'wp-user-frontend' ) . "\r\n\r\n";
-        $message .= $login_url . "\r\n\r\n";
-        $message .= __( 'Thanks', 'wp-user-frontend' );
-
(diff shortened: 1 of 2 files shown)


includes/free/class-login.php
```diff
@@ -1091,4 +1079,64 @@ public function default_wp_login_override() {
             exit;
         }
     }
+
+    /**
+     * Send activated email for user
+     *
+     * @param $user
+     */
+    private function send_activated_email( $user ) {
```
tareq1988, line 1088: The proper function name should be `send_activation_email`

includes/free/class-login.php
```diff
@@ -1091,4 +1079,64 @@ public function default_wp_login_override() {
...
+        $blogname   = wp_specialchars_decode( get_option( 'blogname' ), ENT_QUOTES );
+        $user_email = $user->user_email;
+        $to         = $user_email;
```
tareq1988, line 1091: You don't need to declare variables if you don't need to. Both `$user_email` and `$to` variables aren't needed. Use `$user->user_email` directly.

includes/free/class-login.php
```diff
@@ -1091,4 +1079,64 @@ public function default_wp_login_override() {
...
+    /**
+     * Preapring mail body for activated email
+     *
+     * @param $message
+     * @param $user
+     * @param $blogname
+     *
+     * @return string|string[]
+     */
+    private static function prepare_mail_body( $message, $user, $blogname ) {
```
tareq1988, line 1127: Do we really need to use a static function here? Unless explicitly necessary, don't make it a static function.

tareq1988 (changes requested):

includes/free/class-login.php
```diff
@@ -1091,4 +1079,62 @@ public function default_wp_login_override() {
             exit;
         }
     }
+
+    /**
+     * Send activated email for user
+     *
+     * @param $user
+     */
+    private function send_email_after_activation( $user ) {
```
tareq1988, line 1088: I've asked to name the function as `send_activation_email`, not sure why have you changed it?

zamandevs: I've thought of activation as when it happens i.e when we sent an email for user account activation. but this email is triggered after activation when the admin approves the user. please advise me if i am doing wrong.

tareq1988: Still, keep the name as I suggested.

PR #1221, opened 2021-12-23
fix: Datetime related wrapper function introduced

sapayth:
Some wrapper function compatibility for WP version < 5.3


wpuf-functions.php
```diff
@@ -4012,3 +4012,63 @@ function wpuf_payment_success_page( $data ){
...
+}
+
+/**
+ * Function wp_timezone_string() compatibility for wp version < 5.3
+ *
+ * @since WPUF
+ *
+ * @return string
+ */
+function wpuf_wp_timezone_string() {
```
zamandevs, line 4053: `wpuf_timezone_string` This is proper name.

PR #1222, opened 2021-12-24
fix: Preview option added in user profile

sapayth:
A preview option is added for drafted or unpublished posts in the frontend user profile.
fix [this issue in pro](https://github.com/weDevsOfficial/wpuf-pro/issues/358)

### DIFF
--- a/class/frontend-account.php
+++ b/class/frontend-account.php
@@ -421,6 +421,10 @@ class WPUF_Frontend_Account {
      * @return void
      */
     public function process_user_for_previewing_post( $query ) {
+        if ( current_user_can( 'edit_posts' ) ) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


class/frontend-account.php
```diff
@@ -411,4 +412,40 @@ public function update_profile() {
...
     }
+
+    /**
+     * Check if the user is the post author and give permission for previewing
+     *
+     * @since WPUF
+     *
+     * @return void
+     */
+    public function process_user_for_previewing_post( $query ) {
```
zamandevs, line 423: Better handle admin. admin previewing posts often.
`if ( is_admin() ) return;`

sapayth: > Better handle admin. admin previewing posts often. `if ( is_admin() ) return;`

Bhai as discussed with you, current_user_can( 'edit_posts' ) will check for the admin and other users who has the 'edit_posts' capability

PR #1238, opened 2022-02-01
special character password does not work due to sanitization fixed #1233

### LINKED ISSUE #1233: Backend created special characters passwords are not recognized by WPUF login form during loging in
Backend created special characters passwords are not recognized by WPUF login form during loging in.

**How to reproduce:**

1. Create a user from the backend Users > Add New
2. Password should contain special characters like 0h*;~EtW~:If/g|;&afihOvLucY?k!p1*rCZ"W.$IS1.E'
3. Now, try to log in from WPUF login page/form.
4. You will get the error message "Error: The password you entered for the username is incorrect. Lost your password?"

_Side note:_ It will work fine if you use the same special password throughout the WPUF registration form and login page. So it's only sanitizing while you were performing like the above method.

**Video:**
https://somup.com/c3VZYIZxxO

zamandevs:


includes/free/class-login.php
```diff
@@ -419,8 +419,8 @@ public function process_login() {
                 return;
             }
 
-            $log = isset( $_POST['log'] ) ? sanitize_text_field( wp_unslash( $_POST['log'] ) ) : '';
```
tareq1988, line 422: `wp-login.php` uses this:

```php
if ( isset( $_POST['log'] ) ) {
    $user_login = ( 'incorrect_password' === $errors->get_error_code() || 'empty_password' === $errors->get_error_code() ) ? esc_attr( wp_unslash( $_POST['log'] ) ) : '';
}
```

So use `wp_unslash` on username.

includes/free/class-login.php
```diff
@@ -484,6 +484,14 @@ public function process_login() {
             $secure_cookie = is_ssl() ? true : false;
             $user          = wp_signon( apply_filters( 'wpuf_login_credentials', $creds ), $secure_cookie );
 
+            //try with old implementation, which is wrong but we must support that
+            if ( is_wp_error( $user ) ) {
+                $creds['user_login']    = sanitize_text_field( wp_unslash( $_POST['log'] ) );
+                $creds['user_password'] = sanitize_text_field( wp_unslash( $_POST['pwd'] ) );
```
tareq1988, line 490: I think you also should remove the `sanitize_text_field` from here as it'll change the password.

zamandevs: Users who register through our form and use special characters they need to login this way as we stored their password through sanitize_text_field. this is a second try if he fails through trim ( we guess he might register before through this way )

tareq1988: Okay

PR #1239, opened 2022-02-02
fix: Validation for email and URL

### LINKED ISSUE #1237: Multistep validation not working for: Website URL, Email, and Email address fields
Validation errors for invalid email addresses or invalid URLs aren't working as you move through multi-step registration forms. The (correct format) validation isn't working across any of the following fields:

- Website URL
- E-Mail (WP user email)
- Email address

It means users CAN move through the steps (even when there is an error on the step). But the final form submit is then blocked by the (invisible) field format errors. For users it would be almost impossible to work out where the error is and why it's not working. 

See this working example (site in development): http://cde.bulletserve.net/dev-multi-step-form-test/
Enter "google" instead of "https://www.google.com" in any/all of the URL fields as an example. (I'm running WP 5.9 and the latest versions of WPUF Pro.)

Thanks

sapayth:
Validation added for invalid email and URL formats.
fix #1237

### DIFF
--- a/assets/js/frontend-form.js
+++ b/assets/js/frontend-form.js
@@ -511,6 +511,42 @@
             WP_User_Frontend.removeErrors(self);
             WP_User_Frontend.removeErrorNotice(self);
 
+            /* ======== Validation ======== */
+            var visibleFields = self.find('input[type="url"]:visible, input[type="email"]:visible');
+
+            visibleFields.each(function(i, item) {
+                var input_type = $(item).attr('type');
+                var val = '';
+
+                switch (input_type) {
+                    case 'url':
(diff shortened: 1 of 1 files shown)


assets/js/frontend-form.js
```diff
@@ -511,6 +511,42 @@
...
+
+                                WP_User_Frontend.markError( item,  error_type );
+                            }
+                        }
+                        break;
+                }
+            });
+
             // ===== Validate: Text and Textarea ========
             var required = self.find('[data-required="yes"]:visible');
```
tareq1988, line 551: Instead of creating a new code block for validating only the visible fields, modify this selector and add the logic here instead.

PR #1245, opened 2022-02-25
Fix/add phone field in form builder

sapayth:
Frontend validation and process for Phone field.
Related [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/394)

### DIFF
--- a/assets/js/frontend-form.js
+++ b/assets/js/frontend-form.js
@@ -511,6 +511,19 @@
             WP_User_Frontend.removeErrors(self);
             WP_User_Frontend.removeErrorNotice(self);
 
+            /* ======== Validate: Phone field ======== */
+            var phoneFields = self.find('input[type="text"].wpuf_telephone:visible');
+
+            phoneFields.each(function(i, item) {
+                // Get iti instance
+                var iti = window.intlTelInputGlobals.getInstance($(this)[0]);
+                if ( '' !== iti.getNumber() && ! iti.isValidNumber() ) {
+                    error = true;
+                    error_type = 'validation';
+                    WP_User_Frontend.markError(item, error_type);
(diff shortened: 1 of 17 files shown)


assets/js/frontend-form.js
```diff
@@ -511,6 +511,31 @@
             WP_User_Frontend.removeErrors(self);
             WP_User_Frontend.removeErrorNotice(self);
 
+            /* ======== Validate: Phone field ======== */
```
tareq1988, line 514: You're adding a new section every time for validation, that's not the idea. Initially it was added for required fields checking, now more and more logics are being added. Try to refactor this into something that can be used in all cases instead adding a new logic, like, iterate through each items in the form, and based on the type:

- Check for required fields and validate
- Check for limits
- Check for proper formatting, etc.

assets/js/frontend-form.js
```diff
@@ -959,7 +984,12 @@
                     ed.focus();
                 }
 
-                var data = $(field).closest('li.wpuf-el').find('script').html().match(/(?:bind\()(.*?(?=\)))/)[1].replace(/['"]/g, '').split(',');
+                var data = '';
+
+                if ( $(field).closest('li.wpuf-el').find('script').html() ) {
+                    var fieldScript = $(field).closest('li.wpuf-el').find('script').html();
+                    data = fieldScript.match(/(?:bind\()(.*?(?=\)))/) ? fieldScript.match(/(?:bind\()(.*?(?=\)))/)[1].replace(/['"]/g, '').split(',') : '';
```
tareq1988, line 991: What's happening here, leave a comment in the code.

tareq1988 (changes requested):

PR #1255, opened 2022-03-23
[PRO] Fix/conditional logic not working on metabox

zamandevs:
Companion [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/399)


Chooton838:
Conditional login in Form is working but there are issues and these are:

- While updating a condition of the form, needs to update all of the conditions otherwise all of them become blank

- Condition in multi-select -> User select the multi-select option then undo the condition -> multi-select disappears -> user submits the form with others fields -> But in the backend, form data shows the multi-select value (though it was not selected at the time of submitting the form)

sapayth:
A new PR is created for the PRO version [here](https://github.com/weDevsOfficial/wpuf-pro/pull/479)

PR #1277, opened 2022-05-20
Wrong settings option assign

mmarj:
This form should come from the WPUF > My Account > [Submission Form](https://prnt.sc/gkQd5KNKziMq)

### DIFF
--- a/templates/submit-post.php
+++ b/templates/submit-post.php
@@ -1,7 +1,7 @@
 <?php
 
 $no_form_notice = __( 'No post form assigned yet by the administrator.', 'wp-user-frontend' );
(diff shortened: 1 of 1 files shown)


zamandevs (changes requested):
This should be a fallback. seems the default fallback won't work. so we need to handle both scenario: 
`$selected_form  = wpuf_get_option( 'post_submission_form', 'wpuf_my_account' );
$selected_form  = empty( $selected_form ) || 'publish' !== get_post_status ( intval( $selected_form ) ) ? wpuf_get_option( 'default_post_form', 'wpuf_frontend_posting' ) : $selected_form ;
`

PR #1278, opened 2022-05-24
admin dashboard news and update metabox added

zamandevs:
Details [HERE](https://docs.google.com/document/d/1PsxSIoXtLVwww7MzGHmDy5PE2eCmx3hpWoH1GjgBglk/edit)

### DIFF
--- a/admin/dashboard-metabox.php
+++ b/admin/dashboard-metabox.php
@@ -1,0 +1,177 @@
+<?php
+
+class Dashboard_Metabox {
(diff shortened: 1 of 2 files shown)


Chooton838:
In the Admin dashboard, News and Updates meta box is added successfully, but there are some issues and these are:

- News & Updates URLs are not redirecting to the WPUF site instead, it redirects to the hosted WP site
- News & Updates URLs are concatenated with the URL of the hosted WP site, but it should be the News & Updates URLs of the WPUF site
- “:” is missing after "https" and it makes the URLs invalid

PR #1288, opened 2022-06-30
Custom hook used to load form builder

sapayth:
We were using a dynamic hook depending on the menu title. If the title got translated, the hook doesn't work for some languages(Español, Türkçe, Arabic, etc.) and shows a blank post form.

We introduced 2 hooks `wpuf_load_post_forms` and `wpuf_load_profile_forms` earlier and now using those hooks to load our form builder

### DIFF
--- a/admin/class-admin-settings.php
+++ b/admin/class-admin-settings.php
@@ -137,6 +137,11 @@ class WPUF_Admin_Settings {
      * Fire when post form submenu registered
      */
     public function post_form_menu_action() {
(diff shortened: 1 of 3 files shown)


fofen:
@sapayth Thank you so much, yes I'm using Chinese. I temporarily removed the translation of menu name and rebuilt the .mo file, now it runs smoothly.

sapayth:
updated after `version 4`

PR #1289, opened 2022-06-30
[FREE] fix: frontend form validation

### LINKED ISSUE #1253: Required Product Image field is not behaving as a required one in the WooCommerce Product upload form template.
**Description:**
If Product Image field in the WooCommerce Product upload form template is set required, then it is not working as a required one. If user click the submit button of that form without populating the Product Image field, the form uploads the product. Product can be found in the db without Product image. 

**How to reproduce the issue:**
- Go to WP Admin Panel -> WP user Frontend -> Post Form -> Add Form
- Select the WooCommerce Product form template.
- Make sure that Product Image field is required https://snipboard.io/rMmftV.jpg
- Publish the form to a page.
- Submit a product by using this form from the frontend.
- Don't upload image in the Product Image field.
- It will upload the product without image, although image was required.

**Expected Behaviour:** 
The form should not let the user upload product without the image if Product Image field is required.

**System Status:**
WooCommerce: v6.3.1
WP User Frontend lite: v3.5.27

sapayth:
- Required fields validation fixed for all the fields
- Content restriction fields validation fixed for both word and character
Companion [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/473) in pro

fix #1253


ehsanulabrar:
Issue: The required option for the `Checkbox field` isn’t working as expected. Previously the required fields issue was addressed and sorted for `Upload fields`. This is [the thread](https://wedevs.slack.com/archives/G17RY7NVB/p1665216577568029).

Please have a look at the attached recording performed on the local site after updating the WP User Frontend Pro plugin with the shared (fixed) zip file on the thread.

https://user-images.githubusercontent.com/54323595/201305464-62584a0e-aa48-4a97-810c-a5a814b4c5cf.mov

PR #1297, opened 2022-09-05
fix: [wpuf-registration] short-code encryption update

sapayth:
Admin can set the short-code with `role` attribute. Example: `[wpuf-registration role='editor']`. Any role will work except `administration`. If admin set `[wpuf-registration role='administration']`, it will set the user role to the default user role from `Settings > General Settings > New User Default Role`

Another [testing scenario](https://wedevs.slack.com/archives/G17RY7NVB/p1662374469651519). Needs to test with PHP version 5.6 and PHP 8

### DIFF
--- a/wpuf-functions.php
+++ b/wpuf-functions.php
@@ -3114,10 +3114,12 @@ add_filter( 'display_post_states', 'wpuf_admin_page_states', 10, 2 );
  * Encryption function for various usage
  *
  * @since 2.5.8
+ * @since WPUF param $nonce added
  *
  * @param string $id
+ * @param string $nonce
  *
- * @return string $encoded_id
+ * @return string|bool encoded string or false if encryption failed
  */
 function wpuf_encryption( $id, $nonce = null ) {
     $auth_keys = WPUF_Encryption_Helper::get_encryption_auth_keys();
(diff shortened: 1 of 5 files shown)


wpuf-functions.php
```diff
@@ -3138,18 +3146,42 @@ function wpuf_encryption( $id ) {
...
-    $iv             = substr( hash( 'sha256', $secret_iv ), 0, 16 );
-    $decoded_id     = openssl_decrypt( base64_decode( $id ), $encrypt_method, $key, 0, $iv );
+    $secret_key = $auth_keys['auth_key'];
+    $secret_iv  = ! empty( $nonce ) ? base64_decode( $nonce ) : $auth_keys['auth_salt'];
 
-    return $decoded_id;
+    // should we use sodium_crypto_secretbox_open
+    if ( function_exists( 'sodium_crypto_secretbox_open') ) {
+        try {
+            return sodium_crypto_secretbox_open( $id, $secret_iv, $secret_key );
```
nurul-umbhiya, line 3164: $id should be base64_decoded eg: `sodium_crypto_secretbox_open( base64_decode( $id ), $secret_iv, $secret_key );`

nurul-umbhiya (approved):

PR #1304, opened 2022-10-19
Added Google ReCaptcha to registration form

joebailey26:
I've been getting massive issues with spammers signing up to my WordPress site.

This PR adds ReCaptcha to the registration form in the same way that it is added to the Login form.

You can see this in action here:
https://southamptonfocus.co.uk/login-register-account/


sapayth:
Hello @joebailey26,
WP User Frontend already has Google ReCaptcha for both our `post form` and `registration form`. For editing the `registration form`, you need the Pro version though. [Here](https://wedevs.com/wp-user-frontend-pro/features/) you can see a list of features WPUF Pro is offering.

Issue #484, opened 2018-06-07
get_settings & like_escape is deprecated


FaraziF:
1-get_settings is deprecated since WordPress 2.1.0! get_option()! wp-user-frontend/includes/class-form.php.
2-like_escape is deprecated since WordPress 4.0.0! wpdb::esc_like()!
wp-user-frontend/wpuf-functions.php.

PR #1309, opened 2022-11-16
fix: subscription shouldn’t be editable from user’s profile

sapayth:
The admin should not change the subscription value from the user profile. The values are coming from the subscription package. In the user profile, it should be disabled

fix [#484](https://github.com/weDevsOfficial/wpuf-pro/issues/484) in pro


Rat01047:
Issue-1:
Can still edit date in Admin-Profile

Screenshot given for reference:
image

Issue #495, opened 2018-06-26
Warning thrown in comments


itowhid06:
![image](https://user-images.githubusercontent.com/12055657/41890623-1da6cbc6-7932-11e8-9478-5c65b6bb4103.png)

PR #1318, opened 2022-12-05
fix: delete inside column fields bug

sapayth:
- delete inside column fields bug
- update popup `confirmButton` and `cancelButton` classes as per sweatalert2 new version
- changed `type` into `icon` as per sweatalert2 doc

**Testing instruction**: Please run `grunt release` after pulling the PR. Needs to test on all the pages where we have sweatalert2. Mostly in the form builder, file upload field in the frontend

fix [#495](https://github.com/weDevsOfficial/wpuf-pro/issues/495)


Rat01047:
Issue-1:
`Phone field` and `Address field` > Design is breaking when adding to Column

Scenario:
Adding `Phone field` and `Address field` > Design is breaking when adding to Column

Screenshot given for reference:
https://www.awesomescreenshot.com/image/35101187?key=158fbd16da1f8d17331b45891603ffb4

PR #1362, opened 2023-06-12
Fix Image Deleted by Both Confirmation and Cancel Buttons

Mohamedbenarbia:
Image Deletion Issue in Confirmation Window: Deletion Occurs When Clicking Either the Confirmation or Delete Button

WPUF


sapayth:
fixed in [this PR](https://github.com/weDevsOfficial/wp-user-frontend/pull/1383)

PR #1389, opened 2023-09-13
enhance: repeat field enhancement

### LINKED ISSUE #93: Add field type feature on the repeating field.
Currently, the repeating field is hardcoded as a text field. It would be nice if we can set a field type dropdown. So we can use that to take inputs as dates, time etc.

sapayth:
Currently, the repeating field is hardcoded as a text field. With this enhancement, we could make almost any field into a repeating field.

fix #93


Rubaiyat-E-Mohammad:
Merge conflict @sapayth vai
CleanShot 2025-07-31 at 15 27 20@2x

PR #1419, opened 2024-01-22
fix: default form template not selecting

sapayth:
when selecting a registration form template, in form settings > General > Form Template not automatically selecting.

image


anik-fahmid:
Event calendar selection also working fine.

PR #1435, opened 2024-02-09
fix: subscription posting restriction not working

sapayth:
The post limit was not working for subscription pack. fixed in this PR


Rat01047:
Approving this PR...as posting restriction is now working

But the other options always had issue from the beginning -> which need fixing.

@anik-fahmid

anik-fahmid:
We have to check and decide the options to keep in s**ubscriptions>postings restrictions** as many features seems not working and may not needed. We will decide this soon when working on revamp. We can proceed with the release as many clients issues need to be solved immediately.

PR #1460, opened 2024-06-11
Fix mail body replace text bug

egushinnosuke:
# Bug detail
When sending emails for creating or updating posts, the function prepare_mail_body() in the Frontend_Form_Ajax class sometimes causes the email body to display the output of wp_get_attachment_url() for related posts instead of showing the meta_value.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of email body preparation to correctly manage attachment URLs based on input type.

### DIFF
--- a/includes/Ajax/Frontend_Form_Ajax.php
+++ b/includes/Ajax/Frontend_Form_Ajax.php
@@ -669,8 +669,30 @@ class Frontend_Form_Ajax {
         preg_match_all( '/%custom_([\w-]*)\b%/', $content, $matches );
         [ $search, $replace ] = $matches;
 
+        $args = array(
+            'post_type' => 'wpuf_input',
+            'posts_per_page' => -1
+        );
+        $wpuf_posts = get_posts($args);
+
         if ( $replace ) {
             foreach ( $replace as $index => $meta_key ) {
+                $input_type = '';
+                foreach ($wpuf_posts as $post) {
+                    $post_content = $post->post_content;
+                    $data = unserialize($post_content);
+                    if ($data['name'] == $meta_key){
+                        $input_type = $data['input_type'];
+                        break;
+                    }
+                }
+
+                $use_attachment_url = false;
+                if ($input_type == "image_upload" || $input_type == "file_upload"){
+                    $use_attachment_url = true;
+                }
+                    
+
                 $value = get_post_meta( $post_id, $meta_key, false );
 
                 if ( isset( $value[0] ) && is_array( $value[0] ) ) {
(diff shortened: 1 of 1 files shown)


sapayth:
there could be an optimised approach. can you share the bug details in a video please @egushinnosuke ?

egushinnosuke:
@sapayth please check below
https://gyazo.com/31e6925bffeb918dd99b05c6f12ae03d
PW：bug0708

egushinnosuke:
@anik-fahmid @sapayth 
Please merge this. 
I have to manually rewrite the code every time the plugin is updated.

anik-fahmid:
@egushinnosuke please wait till the next release.

@arifulhoque7 will check and add it in the next RC.

arifulhoque7:
The bug is real: `prepare_mail_body()` detects attachments via `get_post_mime_type( (int) $val )`, so any plain **numeric** custom-field value that happens to match a real post/attachment ID gets wrongly turned into a `wp_get_attachment_url()` in the email. Switching to the field's `input_type` is the right approach. A few things to address before merge:

### Blocking

1. **Scope bug (correctness).** `get_posts(['post_type' => 'wpuf_input', 'posts_per_page' => -1])` loads **every field of every form on the site**, not this post's form. Two forms with a same-named field → wrong `input_type` is picked. It should be scoped to the post's own form.

2. **Reuse the existing API.** `wpuf_get_form_fields( $form_id )` already does exactly this — `get_children` scoped to the form + `maybe_unserialize` + `input_type` with an `isset` guard. No need to reimplement the query/unserialize.

3. **Unsafe `unserialize()`.** Raw `unserialize($post->post_content)` with no guard emits PHP warnings (and `$data['name']` on `false`) when the content isn't the expected shape. Use `maybe_unserialize()` + `isset()`.

4. **`$post` shadowing.** The inner `foreach ($wpuf_posts as $post)` clobbers the outer `$post = get_post( $post_id )`. Rename the loop variable.

5. **Performance.** Per meta-key it re-loops all inputs and unserializes each (O(fields × inputs)). Build a `name => input_type` map once, before the loop.

### Coding standards (WPCS — enforced on this repo)
- Loose `==` → `===` (`$data['name'] == $meta_key`, `$input_type == "image_upload"`).
- `array(...)` → short array `[]`.
- Spacing inside parens: `get_posts( $args )`, `foreach ( $x as $y )`.
- Trailing whitespace on the blank line after the `if` block.
- Build/run the query inside the `if ( $replace )` guard so it's skipped when there are no custom fields.

### Suggested approach
```php
$form_id     = (int) get_post_meta( $post_id, '_wpuf_form_id', true );
$field_types = [];

if ( $form_id ) {
    foreach ( wpuf_get_form_fields( $form_id ) as $field ) {
        if ( ! empty( $field['name'] ) ) {
            $field_types[ $field['name'] ] = isset( $field['input_type'] ) ? $field['input_type'] : '';
        }
    }
}

// inside the loop, per $meta_key:
$use_attachment_url = in_array(
    isset( $field_types[ $meta_key ] ) ? $field_types[ $meta_key ] : '',
    [ 'image_upload', 'file_upload' ],
    true
);
```

This keeps your type-based detection (which is correct) while fixing scope, safety, performance, and standards. Happy to help with the revision.

PR #1523, opened 2025-03-25
deselect form template on dblclick

sapayth:
When a user selects a form template from Post Form > Settings > General > Choose Form Template, they cannot currently deselect it. This PR introduces a deselect option, allowing users to double-click the selected template to remove the selection.

Now, only available form templates are shown. Templates that require a third-party integration will be hidden if the service is not active.

- [ ] Deselect option
- [ ] show only available template

CleanShot 2025-03-25 at 11 01 45


Rubaiyat-E-Mohammad:
Issues got so far:-

1. woocommerce, edd templates are missing in post form settings
2. dblClick not working in registration form settings
@sapayth vai, @anik-fahmid vai

https://github.com/user-attachments/assets/5467c6f5-d0a3-4e61-960b-6fd7d5ccb32b

PR #1541, opened 2025-04-18
Fix: Display 'Pending Review' posts in the frontend

### LINKED ISSUE #1538: Post Submission Status - Pending Review doesn't show up in Posts
Hi team,

I'd like to report that when setting the _Post Forms > Before Post Settings > Post Submission Status_ to **Pending Review**, and then someone adds a post to our site, it doesn't show up in WordPress > Posts.

However, the post does exist, it just cannot be found, nor even searched for. You need the direct link to it. That is a strange bug since it means we'll never know nor see someone's submission.

I have to set it to Draft for it to work. I had it set to draft for many years and it was working fine but last week decided to try the Pending Review status instead.

Hopefully you can resolve this.

arifulhoque7:
Close #1538

### DIFF
--- a/wpuf-functions.php
+++ b/wpuf-functions.php
@@ -5019,7 +5019,7 @@ function wpuf_get_post_form_builder_setting_menu_contents() {
                             'type'      => 'select',
                             'options'   => [
                                 'draft'          => __( 'Draft', 'wp-user-frontend' ),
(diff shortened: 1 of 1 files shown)


sapayth:
already fixed and merged with #1545

PR #1619, opened 2025-07-08
fix:  PPP Cost is Empty and Saved in Post form Payment Settings

arifulhoque7:
Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/883)
## Summary of Changes Made

The "Cost for each additional post after pack limit is reached" field is now mandatory when "Pay-per-post billing when limit exceeds" is enabled. Here's what we implemented:

### 1. **PHP Form Settings** (wpuf-functions.php)
- Added `'required' => true` to the `fallback_ppp_cost` field definition
- Added helpful text explaining when the field is required

### 2. **Frontend JavaScript Validation** (wpuf-form-builder.js)
- Added `validation_error_msg` property to the Vue component data
- Created `validate_form_before_submit()` method that:
  - Checks if payment options are enabled
  - Verifies if "Force Pack Purchase" is selected
  - Confirms if "Pay-per-post billing when limit exceeds" is checked
  - Validates that the cost field is not empty or zero
  - Shows an error message and prevents form saving if validation fails

### 3. **Form Builder Configuration** (form-builder.js)
- Fixed the field type from 'checkbox' to 'number'
- Added `required: true` property
- Set up proper dependency conditions

### 4. **Server-side Validation** (Admin_Form_Builder_Ajax.php)
- Added `validate_fallback_ppp_cost_required()` method for server-side validation
- Integrated the validation into the form saving process
- Returns an error response if validation fails, preventing form save

### How It Works:

1. **When the user tries to save the form**, the JavaScript validation runs first
2. **If all conditions are met** (payments enabled + force pack purchase + fallback PPP enabled), it checks if the cost field has a valid value
3. **If the cost field is empty or zero**, it shows a warning message and prevents saving
4. **If JavaScript is disabled**, the server-side validation catches the same condition and returns an error
5. **The form cannot be saved** until a valid cost is entered when the fallback PPP option is enabled

The validation is now fully functional and prevents users from saving incomplete payment configurations!

iScreen Shoter - 20250708104647917

## Summary by CodeRabbit

* **New Features**
  * Added client-side and server-side validation to ensure a fallback pay-per-post cost is provided when required in payment settings.
  * Improved error feedback by displaying validation errors in warning dialogs before form submission.

* **Bug Fixes**
  * Updated default behavior so that the fallback pay-per-post cost now defaults to zero if not set, instead of two.

* **Documentation**
  * Added help text to clarify when the fallback pay-per-post cost field is required in the form builder UI.

### DIFF
--- a/admin/form-builder/assets/js/form-builder.js
+++ b/admin/form-builder/assets/js/form-builder.js
@@ -532,6 +532,7 @@
             post_title_editing: false,
             isDirty: false,
             shortcodeCopied: false,
+            validation_error_msg: '',
             logoUrl: wpuf_form_builder.asset_url + '/images/wpuf-icon-circle.svg',
             settings_titles: wpuf_form_builder.settings_titles,
             settings_items: wpuf_form_builder.settings_items,
@@ -725,7 +726,7 @@
                 if (_.isFunction(this.validate_form_before_submit) && !this.validate_form_before_submit()) {
 
                     this.warn({
-                        title: 'Incomplete Post Form',
+                        title: 'Post Form Validation Error!',
                         html: this.validation_error_msg,
                         reverseButtons: true,
                         customClass: {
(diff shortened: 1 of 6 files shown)


admin/form-builder/assets/js/form-builder.js
```diff
@@ -813,8 +828,43 @@
                 }
             },
 
+            // Validate form before submitting
+            validate_form_before_submit: function () {
```
sapayth: `validate_form_before_submit` function already exists in `wp-user-frontend/assets/js/wpuf-form-builder-wpuf-forms.js:21`. try to organise all the validation in one place

PR #1700, opened 2025-09-10
Prevent CSS leaks by restricting form-builder styles to WPUF pages only

arifulhoque7:
### **Bug Fix: Prevent WPUF Form Builder CSS from affecting all WP Admin pages**

**Problem**
The `wpuf-admin-form-builder.css` file contained the `@tailwind base` directive, which applies a global CSS reset (e.g., `margin: 0`, `padding: 0`, border resets). Since the stylesheet was being enqueued globally, it caused visual inconsistencies and broke styling on unrelated WordPress admin pages.

**Root Cause**
The CSS was enqueued unconditionally in:

1. `includes/Admin/Posting.php:108` → `wp_enqueue_style('wpuf-admin-form-builder')`
2. `includes/Admin/Forms/Post/Templates/Form_Template.php:85` → same enqueue call

Because these enqueue calls had no page restrictions, Tailwind’s reset styles leaked into every admin page, affecting WordPress core UI, plugins, and themes.

**Investigation Notes**

* This issue was **difficult and time-consuming to debug** because the CSS reset blended with native WordPress styles.
* At first glance, the broken layouts appeared unrelated to WPUF, making it tricky to identify the true source.
* Only after carefully isolating and testing across multiple admin screens was the **actual cause (unscoped Tailwind base styles)** discovered.
* The process took significantly more effort than a normal bugfix due to the subtle and widespread impact.

**Solution**

* Added conditional enqueue logic using `get_current_screen()`.
* Styles are now only loaded when the current admin screen is related to **WPUF** (`$screen->id` or `$screen->base` contains `wpuf`).
* Restored `@tailwind base` in the source file since scoping is safely handled at the enqueue level.

**Testing**

* Verified CSS only loads on WPUF-specific admin screens.
* Checked multiple WordPress admin pages to ensure no unintended style overrides remain.
* Confirmed full functionality of the form builder and backward compatibility.

---

✅ This resolves the **global CSS leakage issue**.
⚡️ Debugging this took much longer than a typical fix because the problem was deeply hidden and only revealed itself after exhaustive analysis.

### DIFF
--- a/includes/Admin/Forms/Post/Templates/Form_Template.php
+++ b/includes/Admin/Forms/Post/Templates/Form_Template.php
@@ -82,7 +82,11 @@ class Form_Template {
             wp_enqueue_style( $deps );
         }
 
-        wp_enqueue_style( 'wpuf-admin-form-builder' );
+        // Only load on WPUF admin pages to prevent CSS leaks
+        $screen = get_current_screen();
+        if ($screen && (strpos($screen->id, 'wpuf') !== false || strpos($screen->base, 'wpuf') !== false)) {
+            wp_enqueue_style( 'wpuf-admin-form-builder' );
+        }
     }
 
     /**
--- a/includes/Admin/Posting.php
+++ b/includes/Admin/Posting.php
@@ -105,7 +105,11 @@ class Posting {
     public function enqueue_script() {
         $api_key = wpuf_get_option( 'gmap_api_key', 'wpuf_general' );
 
-        wp_enqueue_style( 'wpuf-admin-form-builder' );
+        // Only load on WPUF admin pages to prevent CSS leaks
+        $screen = get_current_screen();
+        if ($screen && (strpos($screen->id, 'wpuf') !== false || strpos($screen->base, 'wpuf') !== false)) {
+            wp_enqueue_style( 'wpuf-admin-form-builder' );
+        }
 
         wp_enqueue_style( 'jquery-ui' );
         wp_enqueue_script( 'jquery-ui-slider' );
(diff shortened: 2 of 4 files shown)


includes/Admin/Forms/Post/Templates/Form_Template.php
```diff
@@ -82,7 +82,11 @@ public function enqueue_scripts() {
             wp_enqueue_style( $deps );
         }
 
-        wp_enqueue_style( 'wpuf-admin-form-builder' );
+        // Only load on WPUF admin pages to prevent CSS leaks
+        $screen = get_current_screen();
+        if ($screen && (strpos($screen->id, 'wpuf') !== false || strpos($screen->base, 'wpuf') !== false)) {
```
sapayth: instead of checking the current screen, can we use `wpuf_load_post_forms` hook? `wpuf_load_post_forms` hook fires only on wpuf post form page

includes/Admin/Posting.php
```diff
@@ -105,7 +105,11 @@ public static function init() {
     public function enqueue_script() {
         $api_key = wpuf_get_option( 'gmap_api_key', 'wpuf_general' );
 
-        wp_enqueue_style( 'wpuf-admin-form-builder' );
+        // Only load on WPUF admin pages to prevent CSS leaks
+        $screen = get_current_screen();
+        if ($screen && (strpos($screen->id, 'wpuf') !== false || strpos($screen->base, 'wpuf') !== false)) {
```
sapayth: instead of checking the current screen, can we use `wpuf_load_post_forms` hook? `wpuf_load_post_forms` hook fires only on wpuf post form page

arifulhoque7: done @sapayth vai

PR #1748, opened 2025-11-24
fix: correctly detect ACF taxonomies by checking post_content

### LINKED ISSUE #1672: Unexpected Behavior in Taxonomy Input Field on Post Form v4.1.8
I am experiencing unexpected behavior with the taxonomy input field in the post form of the WordPress User Frontend plugin.

The taxonomy field offers several input types:
- Select
- Multi Select
- Checkbox
- Text Input
- Ajax

There is also a Selection Type setting with the following options:

- Exclude
- Include
- Child of

All input types seem to behave the same way, and the advanced Selection Type options (Exclude, Include, Child of) do not appear to have any effect.

**Expected Behavior**
For example, when I select the **Ajax** input type and set **Child of** to a specific parent category (e.g., **parent C**), the form should only display terms that are children of that parent category (parent C). However, this filtering does not work as expected — **the full list of terms is still shown.**

**Steps to Reproduce**
Go to the post form.
Choose the taxonomy field and set the input type to Ajax.
Set the Selection Type to Child of and specify a parent term (e.g., parent C).
Submit the form and observe the taxonomy field.

**Actual Behavior**
The taxonomy field shows all terms instead of filtering by the specified child of the parent term.

arifulhoque7:
fix: correctly detect ACF taxonomies by checking post_content

The wpuf_is_acf_taxonomy() function was incorrectly checking the post_name
field when querying for ACF-registered taxonomies. ACF stores the actual
taxonomy slug in the serialized post_content under the 'taxonomy' key,
while post_name contains ACF's internal identifier (e.g., 'taxonomy_69242380c35d7').

Changes:
- Modified database query to fetch all acf-taxonomy posts
- Added loop to unserialize post_content and check taxonomy key
- ACF taxonomies now properly detected as free fields
- Removed PRO badge from ACF taxonomy fields in form builder

This ensures ACF custom taxonomies are available in the free version
without requiring PRO upgrade, as they are user-created content.

Close #1672 

## Summary by CodeRabbit

* **New Features**
  * ACF and other non-built-in taxonomies are now included in the free taxonomy set, so they render and can be selected without Pro.
  * Taxonomy visibility/filtering for non‑Pro users updated to preserve these taxonomies when listing or filtering fields.
  * Added utilities to build and expose the free-taxonomy list used by the UI.

* **Chores**
  * Minor array/format refactors for consistency.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/includes/Admin/Forms/Admin_Form_Builder.php
+++ b/includes/Admin/Forms/Admin_Form_Builder.php
@@ -517,21 +517,21 @@ class Admin_Form_Builder {
             return $form_fields;
         }
 
-        // Get free taxonomies (built-in + ACF taxonomies)
+        // Get free taxonomies (built-in + taxonomies for post/page)
         $free_taxonomies = wpuf_get_free_taxonomies();
 
-        // Filter out custom taxonomy fields (excluding ACF taxonomies)
+        // Filter out custom taxonomy fields that are not in the free list
         $filtered_fields = array();
 
         foreach ( $form_fields as $field ) {
-            // Skip custom taxonomy fields when pro is not active (but keep ACF taxonomies)
+            // Skip custom taxonomy fields when pro is not active
             if ( isset( $field['input_type'] ) && $field['input_type'] === 'taxonomy' ) {
                 if ( isset( $field['name'] ) && ! in_array( $field['name'], $free_taxonomies, true ) ) {
-                    continue; // Skip this custom taxonomy field (not ACF)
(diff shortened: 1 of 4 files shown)


sapayth (changes requested):
Remove the ACF related or non-related comments. We are not doing anything special for ACF here.
Check older version for implementation of free taxonomy in our form builder.

sapayth (approved):

PR #1786, opened 2026-01-08
Consolidating or reorganizing TEC integration code

arifulhoque7:
Description: Refactoring TEC integration by removing redundant files and updating templates for better code organization
Related PRO [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/1364)

## Summary by CodeRabbit

* **Changes**
  * Event start/end date/time fields removed from event creation forms.
  * Automatic venue & organizer integration removed.

* **Behavior**
  * Form field data now passes through a filter so pro-only fields can be excluded when Pro is inactive.
  * Form retrieval now returns the filtered fields set.

* **Chores**
  * Free loader registers a filter to remove pro-only fields when Pro is not active.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


Rubaiyat-E-Mohammad:
- [ ] venue and organizer related fields should not be shown here, like event date & event time field @arifulhoque7 vai

image

image



 For Example:

image

includes/Integrations/Events_Calendar/Templates/Event_Form_Template.php
```diff
@@ -75,64 +75,65 @@ public function get_form_fields() {
...
-                'is_meta'    => 'yes',
-                'help'       => '',
-                'width'      => 'large',
-                'format'     => 'yy-mm-dd',
-                'time'       => 'yes',
-                'wpuf_cond'  => $this->conditionals,
-                'id'         => uniqid('wpuf_', true),
-                'is_new'    => true
-            ],
+            // commented out fields for Event Start and End dates as it is pro feature
```
sapayth: remove the codes instead of commenting out

arifulhoque7: done @sapayth vai

includes/Free/Free_Loader.php
```diff
@@ -974,6 +977,37 @@ public function post_form_templates( $integrations ) {
...
+     * Filter out pro-only fields when pro plugin is not active
+     *
+     * @since 4.1.9
+     *
+     * @param array $form_fields The array of form fields
+     * @param int   $form_id     The form ID
+     *
+     * @return array Filtered form fields
+     */
+    public function filter_pro_only_fields( $form_fields, $form_id ) {
```
sapayth: no need to remove the pro fields. our pro fields will only show a notice in the form builder and nothing will be shown where the form render

sapayth (changes requested):
after making the changes, re-check if a form properly submitting without any console or debug error after event calendar form submission

includes/Free/Free_Loader.php
arifulhoque7: done @sapayth vai

PR #1812, opened 2026-02-15
Enable fallback pay-per-post when force_pack set

arifulhoque7:
**Fix: Allow post submission with fallback pay-per-post when subscription post limit is exceeded**. Close [1414](https://github.com/weDevsOfficial/wpuf-pro/issues/1414)

When a form has "Mandatory Subscription" enabled with "Pay-per-post billing when limit exceeds" checked and a fallback cost set, users who have exhausted their subscription post limit were incorrectly blocked with the message "Post Limit Exceeded for your purchased subscription pack." instead of being allowed to submit and pay per post.

**Root cause: Three early return checks in the code were blocking users before the fallback pay-per-post logic could be evaluated:**

Form::is_submission_open() — early return at the top-level post count check
Subscription::force_pack_permission() — wpuf_can_post filter overriding the result back to 'no'
Subscription::force_pack_notice() — displaying "You must purchase a pack" notice instead of the form
Fix: Added a $skip_limit_block condition to all three locations so the early returns are bypassed when force_pack + fallback_cost are both enabled, allowing the existing downstream logic to correctly permit posting with per-post payment.

## Summary by CodeRabbit

* **New Features**
  * Admin users bypass post limits, payment redirects, and subscription notices.

* **UI**
  * Fallback and per-post cost fields switched to numeric inputs with non-negative values and two-decimal precision; dependency input type updated.

* **Bug Fixes**
  * Consistent decimal handling for monetary values; improved interaction between pack enforcement and fallback pay-per-post with skip logic; subscription error checks now return results correctly.


Rubaiyat-E-Mohammad:
As a user, I see the wrong pay per post after post limit exceeded charge showing. 
Also, previous exists for admin end @arifulhoque7 vai

image

image

arifulhoque7:
> As a user, I see the wrong pay per post after post limit exceeded charge showing.

fixed @Rubaiyat-E-Mohammad vai

Rubaiyat-E-Mohammad:
> > As a user, I see the wrong pay per post after post limit exceeded charge showing.
> 
> fixed @Rubaiyat-E-Mohammad vai

Not solved @arifulhoque7 vai

Still getting same error
image

Pay-per-post payment is not being required

https://github.com/user-attachments/assets/49ef9a53-ee68-45af-a230-cbc282e94d95

arifulhoque7:
> Pay-per-post payment is not being required

Please Check it now , it should work @Rubaiyat-E-Mohammad vai

Rubaiyat-E-Mohammad:
> > Pay-per-post payment is not being required
> 
> Please Check it now , it should work @Rubaiyat-E-Mohammad vai

Showing wrong amount issue not solved vai @arifulhoque7 vai

image
image

sapayth (changes requested):
## Critical Issues

### 1. Missing return fix is correct but reveals a deeper problem

`class/subscription.php:1219` and `includes/Admin/Subscription.php:1284` — `has_user_error()` was silently discarding the result. This means every call to `has_user_error()` has been evaluating as falsy since the function was written.

The fix is correct, but this means enabling the return value will now change behavior at every callsite, not just the ones this PR targets. This needs careful E2E testing — fixing this bug might expose other logic that was accidentally relying on `has_user_error()` always returning null/falsy.

---

### 2. Condition in `add_post_info()` elseif may be too narrow

Both `class/subscription.php` and `includes/Admin/Subscription.php` change:

**Before:**

```php
} elseif ( self::has_user_error( $form_settings ) || ( $payment_enabled && ... ) ) {
```

**After:**

```php
} elseif ( $payment_enabled && $force_pack && $form->is_enabled_fallback_cost() && ... ) {
```

The `self::has_user_error()` OR branch is completely removed. Now that `has_user_error()` actually returns a value (fix #1), there may be user-error scenarios that should still show the fallback cost info block but won't reach this branch anymore. Worth verifying there isn't a case where a user has an error, `force_pack` is off, but fallback cost should still display.

---

## Moderate Issues

### 3. `$form` variable shadowing in `add_post_info`

In `includes/Admin/Subscription.php`, the elseif block at ~line 1122 still has `$form = new Form( $form_id );` even though `$form` is already instantiated earlier in the method. The `class/subscription.php` version correctly removes this redundant instantiation, but the `includes/Admin/Subscription.php` version keeps it. This should be cleaned up for consistency.

---

### 4. New early return in `class/subscription.php` `force_pack_permission` changes behavior beyond the stated fix

Lines 1162–1165 in the diff add:

```php
if ( $current_user->subscription()->current_pack_id() && ! $has_post_count && ! $skip_limit_block ) {
    return 'no';
}
```

This early return didn't exist in the legacy `class/subscription.php` before (it only existed in `includes/Admin/Subscription.php`). While this makes the two files more consistent, it changes behavior for users who have a pack with no post count remaining and `$skip_limit_block` is false — they'll now be blocked before reaching the existing `is_user_logged_in()` logic. This is probably fine but should be called out.

---

### 5. Admin bypass scope is broad

The admin bypass in `Form.php::is_submission_open()` returns `['yes', '']` before ANY other check. This means admins bypass post-locking, guest post restrictions, and all payment logic — not just subscription limits. If the intent is only to bypass subscription/payment restrictions, the check should be placed after the post-lock check.

---

## Minor Issues

### 6. Duplicate JS files

Both `admin/form-builder/assets/js/form-builder.js` and `assets/js/wpuf-form-builder.js` have the same checkbox → number type fix. Is one the source and the other compiled? If so, only the source should be edited and the compiled file should be generated via build.

---

### 7. Swapped comment labels are fixed — good catch

The `//fallback cost enabled` / `//fallback cost disabled` comments were swapped in both files. The PR corrects them.

---

### 8. Float casting and input constraints are good

`(int)` → `(float)` for monetary values, `step="0.01"` `min="0"` on inputs — both correct improvements.

---

## Summary

| Category                      | Verdict                                              |
|-------------------------------|------------------------------------------------------|
| Bug fix logic                 | Correct approach, well-reasoned                     |
| has_user_error return fix     | Critical fix, but needs broader regression testing  |
| Admin bypasses                | Work, but broader than necessary                    |
| Code duplication (legacy/modern) | Consistent, expected given architecture          |
| Type safety (float/input)     | Good                                                |

arifulhoque7:
fixed @sapayth vai

sapayth (changes requested):
`has_user_error()` return semantics change has wide blast radius

  The method was always silently returning null (void). The fix adds the missing return, which is the right thing to do — but it changes the truthiness of this call at six other
  pre-existing callsites that this PR leaves untouched:

  - class/subscription.php:632, 1056, 1127
  - includes/Admin/Subscription.php:656, 1095, 1185

  Before this PR those branches were effectively dead (always falsy). After, they'll fire whenever a subscription error exists. At least two of those callsites are if (
  self::has_user_error(...) ) { echo error_message; } paths that were previously unreachable.

  Fix: Walk each callsite and confirm the new behavior is desired. Specifically check: does has_error() return WP_Error in success states, or strictly true/false/WP_Error? If it can
  return a WP_Error object, any if ( self::has_user_error() ) will now be truthy even when there is no real error unless the object itself is falsy.
  Location: class/subscription.php:1246, includes/Admin/Subscription.php:1303

PR #1251, opened 2022-03-09
Fix/add data removing option

sapayth:
In User Frontend > Tools a new setting is introduced to delete some data upon plugin deletion. When you delete the WP User Frontend version, it will also delete all the data related to WP User Frontend Pro as well. This won't happen when the plugins are deactivated.

1. Delete Plugin Settings on Uninstall
2. Delete Database Tables on Uninstall
3. Delete WPUF Forms on Uninstall
4. Delete WPUF Pages on Uninstall

Settings that will be removed are:
all options containing wpuf_

DB tables that will be removed are:
1. wpuf_transaction
2. wpuf_subscribers
3. wpuf_message
4. wpuf_activity

All pages containing wpuf shortcode

## Summary by CodeRabbit

* **New Features**
  * Uninstall settings panel with persistent toggles (delete settings, database, forms, pages) and a caution block explaining data removal.
  * Form export to downloadable JSON for all or selected forms (including registration forms).
  * Form import via JSON uploader with progress/feedback UI.
  * Admin Tools page with maintenance actions and bulk form management/export/import options.

PR #1873, opened 2026-06-16
Fix/add data removing option (recover #1251)

arifulhoque7:
Recovered from #1251 (originally by @sapayth), rebased onto `develop` with merge conflicts resolved. Functionality preserved.

Supersedes #1251.

## New Features

* Added admin Tools page featuring export and import capabilities for forms, allowing users to backup and transfer form configurations as JSON files.
* Introduced uninstall settings section in the Tools admin page, enabling users to selectively choose which plugin data (settings, database tables, forms, and pages) to remove during plugin uninstallation.
* Added automated cleanup routine that handles plugin data removal based on user-selected uninstall preferences.


arifulhoque7:
Review fixes pushed:

- `uninstall.php` — the `LIKE` deletes (options + postmeta) and the WPUF-page lookup now use `$wpdb->prepare()` + `$wpdb->esc_like()` instead of raw interpolated SQL. Cleanup breadth kept (still matches `wpuf_`/`_wpuf_` keys) so no orphaned data, but the wildcard `_` is now escaped and the query is injection-safe.
- `admin/class-tools.php` (export_to_json) — moved `header()` calls **before** `echo` to avoid "headers already sent", switched to `wp_json_encode()` and `gmdate()`.
- `import_json_file` — guards against invalid/non-array JSON before the `foreach` (returns a `WP_Error` instead of warning).

Note: the destructive page delete (`wp_delete_post(..., true)`) is intentional and gated behind the opt-in "Delete Pages" uninstall setting + the `[wpuf` shortcode match — left as designed, just made the query safe.

PR #1825, opened 2026-03-04
Feature/user directory elementor widget

sapayth:
depends on [#1778](https://github.com/weDevsOfficial/wp-user-frontend/pull/1778)

closes [#1410](https://github.com/weDevsOfficial/wpuf-pro/issues/1410), closes [#1445](https://github.com/weDevsOfficial/wpuf-pro/issues/1445), closes [#1446](https://github.com/weDevsOfficial/wpuf-pro/issues/1446)

## Summary

This PR adds an **Elementor widget** so the directory can be embedded anywhere using the Elementor page builder — no shortcodes required.

## What's New

- **Elementor widget:** A dedicated "User Directory" widget is available in Elementor's widget panel, allowing full visual control over placement and styling.

## Technical Notes

- New Elementor widget class: `WeDevs\Wpuf\Integrations\Elementor\User_Directory_Widget`

## Release Notes

* **New Features**
  * Added Elementor User Directory widget with customizable controls for search, filters, user cards, typography, and pagination
  * Enabled pretty URL support for pages using the widget
  * Implemented editor preview for User Directory widget in Elementor

PR #1879, opened 2026-06-16
Feature/user directory elementor widget (recover #1825)

arifulhoque7:
Recovered from #1825 (originally by @sapayth), rebased onto `develop` with merge conflicts resolved. Functionality preserved.

Supersedes #1825.


arifulhoque7:
Review fix pushed: `register_elementor_page_rewrite_rules()` ran a `get_posts(-1)` + `_elementor_data` scan over **all** pages on every front-end `init` whenever no flagged page existed. Now gated behind a daily transient (`wpuf_ud_elementor_backfill_done`). The meta flag is still written on page save (the fast `meta_query` path), and the transient is cleared on save so a newly-added widget page is re-detected. `flush_rewrite_rules()` now only runs inside that one-time back-fill or on editor save, not on normal front-end requests.

PR #1852, opened 2026-04-29
feature: 2fa with authenticator apps

sapayth:
closes [#1506](https://github.com/weDevsOfficial/wpuf-pro/issues/1506)
closes [#1507](https://github.com/weDevsOfficial/wpuf-pro/issues/1507)

## Summary

Adds first-class **Two-Factor Authentication** to WP User Frontend's free edition. Users can secure their accounts with a standards-compliant authenticator app (Google Authenticator, 1Password, Authy, etc.) and are challenged for a 6-digit code on the WPUF login form after their password is verified.

The feature is **off by default** and is gated behind a new admin settings section (**WPUF Settings → 2FA**). Once enabled, an authenticated user gets a new **Security** tab inside `[wpuf_account]` to enroll, regenerate, or disable their authenticator. Admins get a **Reset 2FA** action on `Users → Edit User` for lockout recovery, with an audit trail.

The MVP architecture is multi-method aware (TOTP today; Email OTP / SMS OTP can plug in later via `wpuf_2fa_register_methods`) but ships only TOTP.

Spec: [`docs/totp.md`](docs/totp.md).

---

### User-facing
- New **Security** tab in `[wpuf_account]` (only visible when 2FA is enabled globally).
- Enrollment flow with QR code, manual entry key, and 6-digit confirmation.
- **Regenerate secret** and **Disable 2FA** actions, both re-prompting for password + current code.
- Login challenge on the **WPUF login form** — credentials in step 1, code in step 2; the auth cookie is never set until the code is verified.
- Friendly error states (`That code didn't match.`, `Your sign-in session expired.`, `Too many attempts.`) instead of generic WP errors.

### Admin-facing
- New settings section **WPUF Settings → 2FA** with:
  - `Enable Two-Factor Authentication` toggle
  - `2FA Methods` selector (card grid; TOTP today)
  - `App Display Name` — issuer label shown inside the user's authenticator app, defaults to the site name
- New section on `Users → Edit User` showing enrollment status, recent activity (audit log), and a **Reset 2FA** button (administrators only, capability `manage_options`).

### Architecture / extensibility
- New namespace `WeDevs\Wpuf\TwoFactor\` (`includes/TwoFactor/`):
  - `Two_Factor_Manager` — service container, accessed as `wpuf()->two_factor`
  - `TOTP_Method` implements a generic `Method_Interface`, registered through `Method_Registry` so future methods slot in without rewriting the challenge controller
[...]

PR #1885, opened 2026-06-16
feature: 2fa with authenticator apps (recover #1852)

arifulhoque7:
Recovered from #1852 (originally by @sapayth), rebased onto `develop` with merge conflicts resolved. Functionality preserved.

Supersedes #1852.


arifulhoque7:
Review: 2FA security posture is strong (AES-256-GCM secrets via HKDF, hashed issued codes, TOTP replay protection, rate limiting, single-use 5-min login tokens, `wp_safe_redirect`, password not sanitized before `wp_check_password`, nonce + `is_user_logged_in` on all AJAX, admin reset gated on `manage_options`). No blocking code issues.

Two items to address before release (not code-fixed here as they're product/migration decisions):
1. **Turnstile setting moved** `enable_turnstile`/`wpuf_general` → `login_form_turnstile`/`wpuf_profile`. Existing installs with login Turnstile enabled will **silently lose** it. Please ship a settings migration (or document the re-enable step).
2. **PHP minimum bumped 5.6 → 7.4** for the free plugin (TOTP uses typed properties / nullable returns). Confirm this is an intended requirement bump and note it in the changelog / readme `Requires PHP`.

Minor (multisite): `Admin_User_Profile::handle_reset` checks `manage_options` only; could also check `current_user_can('edit_user', $target_user_id)`.

PR #1900, opened 2026-07-03
Fix: guard method_exists() against null $wpuf_field to prevent PHP 8+ fatal error

### LINKED ISSUE #1832: Incompatibility with php8.0+
Call to php function `method_exists()` throws an error if first argument is null. This call, in `FieldableTrait`, is not secured.

Upgrading to php8.5 with the latest version of this plugin (free version - 4.2.11), and latest WP version (6.9.4) caused a fatal error on the page where the plugin is used.

dev-alamin:
Fixes #1832

## Problem

In both `FieldableTrait::prepare_meta_fields()` and `WPUF_Frontend_Render_Form::prepare_meta_fields()`, `$wpuf_field` is resolved via:

```php
$wpuf_field = wpuf()->fields->get_field( $value['template'] );
```

When `get_field()` returns `null` — e.g. the submitted field's `template` no longer matches a registered field type (deprecated/removed field, or a Pro-only field type referenced while Pro is inactive) — the next line calls:

```php
method_exists( $wpuf_field, 'sanitize_field_data' )
```

`method_exists()` requires its first parameter to be `object|string`. Passing `null` throws a `TypeError` on PHP 8.0+ (PHP 7.x only warned and returned `false`), so any frontend form submission hitting this path with an unresolvable field type currently white-screens for the site visitor.

## Fix

Add a truthy check on `$wpuf_field` before calling `method_exists()`, matching the guard style already used elsewhere in the codebase (`wpuf-functions.php:1094` already does `! empty( $wpuf_field ) && method_exists(...)`):

```php
if ( isset( $posted_field_data ) && $wpuf_field && method_exists( $wpuf_field, 'sanitize_field_data' ) ) {
```

Applied to both occurrences of the identical pattern:
- `includes/Traits/FieldableTrait.php`
- `includes/class-frontend-render-form.php`

## Testing

- Traced both call sites — identical unguarded pattern, while `wpuf-functions.php` already guards correctly.
- Reproduced the fatal on PHP 8 by submitting a form with a field `template` that doesn't resolve; confirmed it now falls through gracefully to the `elseif`/`else` branches instead of throwing.
- 2-line diff, no behavior change on the happy path — low risk.

## Notes for reviewers

Didn't touch `wpuf-functions.php:1094` since it's already correctly guarded. Happy to add a regression test if there's a preferred location/harness for `FieldableTrait` tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form field handling to avoid errors when field data is missing or invalid.
  * Input cleaning now only runs when a valid field instance is available, keeping normal behavior unchanged for valid fields.

### DIFF
--- a/includes/Traits/FieldableTrait.php
+++ b/includes/Traits/FieldableTrait.php
@@ -837,7 +837,7 @@ trait FieldableTrait {
             $wpuf_field = wpuf()->fields->get_field( $value['template'] );
             $posted_field_data = isset( $post_data[ $value['name'] ] ) ? $post_data[ $value['name'] ] : null;
 
-            if ( isset( $posted_field_data ) && method_exists( $wpuf_field, 'sanitize_field_data' ) ) {
+            if ( isset( $posted_field_data ) && $wpuf_field && method_exists( $wpuf_field, 'sanitize_field_data' ) ) {
                 $meta_key_value[ $value['name'] ] = $wpuf_field->sanitize_field_data( $posted_field_data, $value );
                 continue;
             } elseif ( isset( $post_data[ $value['name'] ] ) && is_array( $post_data[ $value['name'] ] ) ) {


arifulhoque7:
Already shipped in **v4.3.9**.

`includes/Traits/FieldableTrait.php:843` on `develop` already carries the exact guard this PR proposes:

```php
if ( isset( $posted_field_data ) && $wpuf_field && method_exists( $wpuf_field, 'sanitize_field_data' ) ) {
```

Released as *"Fatal error on PHP 8 when a form contains a field from a deactivated module"* and #1832 is closed. Closing as duplicate — thanks for the report and the fix.

**Note for follow-up:** the PR description also mentions `WPUF_Frontend_Render_Form::prepare_meta_fields()`, but the diff never touched it. `includes/class-frontend-render-form.php:754` still calls `method_exists( $wpuf_field, ... )` with no null guard, so the same fatal is still reachable through that path. Worth a fresh PR against `develop` for that one line alone.

Issue #1797, opened 2026-01-24
Can't translate two language strings in the login process


Anticosti:
Using Version 4.2.6 (Free + Pro)
I noticed two language strings in the login process that I can't get translated withe the **Loco Translate** plugin:

- Enter
- Don't have an account?

<img width="1841" height="549" alt="Image" src="https://github.com/user-attachments/assets/26d39ca4-7560-4778-b075-5e27fa3b56cd" />

PR #1901, opened 2026-07-03
Fix/untranslatable login register string

dev-alamin:
## Fix – "Don't have an account?" not translatable via Loco Translate / .po tools

Partially addresses #1797

### Problem

`Simple_Login::get_action_links()` builds two similar auth links. The "login" link
correctly wraps its entire phrase in `__()`:

```php
sprintf(
    __( 'Already have an account? <a href="%1$s">%2$s</a>', 'wp-user-frontend' ),
    esc_url( $this->get_action_url( 'login' ) ),
    esc_html( __( 'Login', 'wp-user-frontend' ) )
);
```

The "register" link does not follow the same pattern — "Don't have an account? " is
a raw PHP string embedded directly in the `sprintf()` template, never passed through
`__()`/`_e()`:

```php
$links[] = sprintf( '%sDon\'t have an account? %s<a href="%s">%s</a>', '<span class="wpuf-register-text">', '</span>', $this->get_action_url( 'register' ), __( 'Register', 'wp-user-frontend' ) );
```

Translation tools (Loco Translate, Poedit, `wp i18n make-pot`) build their string
lists by scanning source for gettext function calls. A string outside one of those
calls is invisible to them regardless of `.po`/`.mo` state — which matches the
report and what I found checking the shipped `languages/*.po` files (the phrase
isn't in any of them).

### Fix

Wrap the full phrase in `__()`, following the same pattern already used for the
login link two branches above it, and add `esc_url()`/`esc_html()` on the dynamic
parts for consistency with that same sibling code:

```php
$links[] =
        sprintf(
            /* translators: %1$s: opening span tag, %2$s: closing span tag, %3$s: register URL, %4$s: register text */
            __( '%1$sDon\'t have an account?%2$s <a href="%3$s">%4$s</a>', 'wp-user-frontend' ),
            '<span class="wpuf-register-text">',
            '</span>',
            esc_url( $this->get_action_url( 'register' ) ),
            esc_html( __( 'Register', 'wp-user-frontend' ) )
        );
```

Only whitespace placement changed cosmetically (the space between the question and
the closing `</span>` now sits outside the span instead of inside it) — no visible
or functional change to markup structure or styling.

### Scope note for reviewers

The report lists two untranslatable strings: "Enter" and "Don't have an account?".
This PR only fixes the latter. I searched the entire free plugin (PHP templates,
widgets, and all built JS bundles) for a user-facing "Enter" string anywhere in the
login flow and couldn't find one — the only occurrence of "Enter" in the codebase is
a keyboard-event key comparison in the User Directory admin screen, unrelated to
login. My best guess is that string belongs to WP User Frontend Pro's code, which
isn't in this repository, so I can't verify or fix it here. Flagging this so it
doesn't get lost — happy to take a look if someone can confirm where "Enter" is
rendered.

### Testing

- Confirmed the fixed string is now discoverable by grepping for `__(` calls (what
  Loco Translate's parser and `wp i18n make-pot` rely on).
- Verified the login link `Already have an account? ...` continues to render
  unchanged — no behavior change to sibling code.
- 1-line-of-behavior diff (wrapping + escaping only); no change to the rendered
  HTML structure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the registration prompt so the “Register” link is fully translatable and displays correctly in more languages.
  * Added extra safety when processing submitted field values, helping avoid issues when a field isn’t available.

### DIFF
--- a/includes/Free/Simple_Login.php
+++ b/includes/Free/Simple_Login.php
@@ -330,7 +330,14 @@ class Simple_Login {
         }
 
         if ( $args['register'] && get_option( 'users_can_register' ) ) {
-            $links[] = sprintf( '%sDon\'t have an account? %s<a href="%s">%s</a>', '<span class="wpuf-register-text">', '</span>', $this->get_action_url( 'register' ), __( 'Register', 'wp-user-frontend' ) );
+            $links[] = sprintf(
+                /* translators: %1$s: opening span tag, %2$s: closing span tag, %3$s: register URL, %4$s: register text */
+                __( '%1$sDon\'t have an account?%2$s <a href="%3$s">%4$s</a>', 'wp-user-frontend' ),
+                '<span class="wpuf-register-text">',
+                '</span>',
+                esc_url( $this->get_action_url( 'register' ) ),
+                esc_html( __( 'Register', 'wp-user-frontend' ) )
+            );
         }
 
         if ( $args['lostpassword'] ) {
--- a/includes/Traits/FieldableTrait.php
+++ b/includes/Traits/FieldableTrait.php
@@ -837,7 +837,7 @@ trait FieldableTrait {
             $wpuf_field = wpuf()->fields->get_field( $value['template'] );
             $posted_field_data = isset( $post_data[ $value['name'] ] ) ? $post_data[ $value['name'] ] : null;
 
-            if ( isset( $posted_field_data ) && method_exists( $wpuf_field, 'sanitize_field_data' ) ) {
+            if ( isset( $posted_field_data ) && $wpuf_field && method_exists( $wpuf_field, 'sanitize_field_data' ) ) {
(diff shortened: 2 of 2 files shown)


arifulhoque7:
Both hunks are already on `develop`, shipped in **v4.3.9**.

- **`includes/Free/Simple_Login.php`** — the register link is translatable now, via a slightly different implementation that also escapes the URL and the label (`Simple_Login.php:335`):
  ```php
  esc_html__( "Don't have an account?", 'wp-user-frontend' ),
  ```
  Released as *"Registration login form 'Don't have an account?' text is now translatable"*.
- **`includes/Traits/FieldableTrait.php`** — the `$wpuf_field` null guard landed at line 843 (same change as #1900).

Closing as duplicate. #1797 stays **open** — this only partially addressed it, and the remaining untranslatable strings there still need doing. Thanks for catching it.

PR #1947, opened 2026-08-21
[Revamp] Guided onboarding wizard

anik-fahmid:
Implements the onboarding revamp described in weDevsOfficial/wpuf-pro#694.

The code sits in the free plugin because that is where the setup flow, the pages installer and the settings it writes all live. Pro is detected at runtime, so the Pro-only parts (registration forms, Stripe) render correctly when Pro is active.

## What this adds

A standalone wizard at `wp-admin/index.php?page=wpuf-onboarding`, opened from a new **Onboarding** box at the top of User Frontend › Tools. It renders as its own document and follows the User Directory design system (emerald, 24px step markers, 42px controls).

**Steps:** what you need → post form → registration → settings → plugins → ready

- **What you need** — four cards: frontend posting, registration & login, user directory, payments. Whatever is unticked has its whole step removed from the flow, dropped from the rail and from the final checklist, and its dependent controls hidden in the settings step. The directory has no step of its own; ticking it activates the module.
- **Post form** — pick a template, and choose whether authors can edit or delete their own posts.
- **Registration** — choose or create the login and registration pages; picking an existing page adds the shortcode to it rather than replacing anything. Pages already holding a form are labelled in the dropdown.
- **Settings** — install the UF pages, hide the admin bar, add a logout link, enable payments and pick gateways (bank ticked, PayPal free, Stripe shown with the Pro badge and upgrade link when the gateway is not registered).
- **Plugins** — lists only companion plugins that are not running yet, and the whole step drops out when there is nothing left to install. Comes back if one is later deactivated.
- **Ready** — checklist of what got configured, each row linking to the screen that owns it, plus the diagnostics opt-in.

Progress is stored, so the Tools button offers *Resume* mid-flow and *Start* again once a run is finished.

## Also changed

`Admin_Installer::init_pages()` returned to the caller only for the legacy setup wizard and otherwise redirected to Settings and exited. It now recognises the onboarding page too, so installing pages no longer ends the flow.

## Notes for review

- The legacy 3-step `Setup_Wizard` is untouched and still owns the activation redirect, so a fresh install still lands there. Switching that over is a follow-up, along with setting the `wpuf_setup_wizard` guard when this wizard finishes.
- Companion plugins install on submit rather than per card; four fresh downloads in one request is worth a look on slow hosts.
- Tested on a local site with Pro active, on both the picked-everything and picked-one paths.


anik-fahmid:
@arifulhoque7 could you review this one? It implements the onboarding revamp from weDevsOfficial/wpuf-pro#694.

Worth a closer look at two things: `Admin_Installer::init_pages()` now returns instead of redirecting when the wizard is the caller, and the plugins step installs on submit rather than per card.

arifulhoque7:
Picking this up to carry it the rest of the way, @anik-fahmid — thanks for the groundwork. Moved to my fork as #1963 with your nine commits intact and rebased onto current `develop` (this branch was 39 behind, including the 4.3.11 security follow-up).

Closing this one so review happens in a single place. Everything outstanding is tracked on #1963 and weDevsOfficial/wpuf-pro#694.

PR #1954, opened 2026-08-28
fix(e2e): category/tag validation selector (td vs th primary column)

arifulhoque7:
## Problem
Setup phase LS0016/LS0017 (add post categories / tags) hung for 3 minutes and failed. Root cause: the validation selector was `//tbody[@id="the-list"]//tr//td//strong//a`, but WordPress renders the term name in the list's **primary column as a `<th scope="row">`, not a `<td>`**. So the locator never matched — the term was created server-side, but the assertion waited until the test timeout, and the loop never reached the second term.

This blocked the suite from ever getting past LS0016 locally.

## Fix
- Drop the `//td` segment so the row-title link matches in a `th` or `td` (`validateCategory`, `validateTag`).
- `createPostCategories`: reload the categories screen per term (the add form falls back to a full-page POST when its AJAX handler isn't bound, so reusing the prior form races the reload) and wait on the row with `.first()` + a bounded timeout, so a real failure surfaces in seconds instead of a 3-minute hang.

## Result
Setup now passes **LS0016–LS0028** (categories, tags, maps, recaptcha, turnstile, modules on/off, gateways, AI). Verified on a fresh wp-env.

No product code touched — e2e test-suite only.

https://claude.ai/code/session_019KGP9sp935QBeu32pGt5WY

### DIFF
--- a/tests/e2e/pages/selectors.ts
+++ b/tests/e2e/pages/selectors.ts
@@ -232,14 +232,14 @@ export const Selectors = {
             clickCategoryMenu: '//a[normalize-space()="Categories"]',
             addNewCategory: '//input[@id="tag-name"]',
             submitCategory: '//input[@id="submit"]',
-            validateCategory: (categoryName: string) => `//tbody[@id="the-list"]//tr//td//strong//a[normalize-space()="${categoryName}"]`,
+            validateCategory: (categoryName: string) => `//tbody[@id="the-list"]//tr//strong//a[normalize-space()="${categoryName}"]`,
         },
 
         tags: {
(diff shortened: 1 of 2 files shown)


arifulhoque7:
Tracking issue: weDevsOfficial/wpuf-pro#1686. Root cause: the `validateCategory`/`validateTag` selectors targeted `//td//strong//a`, but WordPress renders the term name in the list's primary column as a `<th scope="row">` — so the locator never matched and LS0016/LS0017 hung until the test timeout. Regression-guard comment follow-up: #1955.

PR #1962, opened 2026-08-31
fix(security): 4.3.11 follow-up — draft gate bypass, registration session fixation, role hardening

arifulhoque7:
Closes weDevsOfficial/wpuf-pro#1688

Follow-up to the researchers re-testing our 4.3.11 build. I verified every one of their claims against the code first — all four are accurate. This PR fixes the three that are still open in code; the fourth (CVE-2026-75823/75824) was already fixed in `dc3ac019` but **missed the 4.3.11 tag by 40 minutes**, so it just needs to ship.

## 1. CVE-2026-17563 — subscription gate bypassed through the draft path
*Reported by Charles Vosburgh*

The 4.3.11 gate keyed on the **shape of the request**:

```php
$is_new_submission = ! isset( $_POST['post_id'] ) || 'new' === $wpuf_form_status;
```

so carrying a `post_id` and omitting `wpuf_form_status` made the handler treat a brand-new submission as an edit and skip the gate — while the update branch still published the post from the form's `edit_post_status`. `draft_post()` had no gate at all, which is where the usable `post_id` came from.

**Fix**
- `draft_post()` runs the same `is_submission_open()` / `wpuf_can_post` gate the renderer applies in `add_post_shortcode()` — **for new drafts only**.
- `submit_post()` decides new-vs-edit from **server state**: `draft_post()` flags a draft `_wpuf_draft_pending`, `submit_post()` clears it once the submission completes.

## 2. CVE-2026-17564 — registration session fixation
*Reported by Yaswanth Reddy Sunkara*

Their diagnosis is the important part, and it is right: **the nonce cannot authenticate this request.** `wp_nonce_field()` on a template that only renders for logged-out visitors mints the nonce at user id 0 with an empty session token, so every anonymous visitor is served the same value and it stays valid for roughly a day. Fetch the page once with no cookies and it can be replayed cross-site. Branching on `wp_verify_nonce()` in 4.3.11 was the right change but does not close this.

**Fix:** `process_registration()` returns early when `is_user_logged_in()`. The registration form only ever renders logged-out, so an authenticated request did not come from it — and this is what actually stops the attack, because the autologin below can no longer clear a victim session and re-issue it for the account it just created.

Their other finding here is confirmed fixed: `wp_safe_redirect()` holds against protocol-relative hosts, embedded authorities and backslash-mangled schemes.

## 3. CVE-2026-75823 — role tampering hardening
*Reported by Murad Akhmedov*

`dc3ac019` (already on develop) authenticates the encrypted role blob, which stops tampering. This adds the defence-in-depth they asked for: assignment now refuses any role holding **site-control capabilities**, instead of only the literal `administrator` slug.

Deliberately a **capability check, not a role allow-list** — an allow-list would break every marketplace install.

## Preserving existing behaviour

This is the part I was most careful about; all three guards are additive and none changes a setting's meaning.

**Role matrix** — verified against every role registered on a Dokan + WooCommerce install:

| role | assignable |
|---|---|
| `administrator` | **no** |
| `editor`, `author`, `contributor`, `subscriber` | yes |
| `shop_manager`, `shop_vendor`, `shop_worker`, `shop_accountant`, `customer` | yes |
| `vendor`, `seller`, `pending_vendor` (Dokan) | yes |
| unknown / non-existent role | no (falls back to `default_role`) |

Filterable via `wpuf_registration_forbidden_capabilities`.

**Draft matrix** — the gate must never catch a genuine edit:

| case | gated |
|---|---|
| brand-new submission (no `post_id`) | yes |
| unsubmitted draft (`_wpuf_draft_pending`) | yes |
| post that already completed a submission | **no** |
| legacy post/draft created before this flag existed | **no** |
| the same draft, after its submission completes | **no** |

So a user whose pack has expired can still edit content they already own, and existing sites see no change on upgrade — posts without the flag are treated as edits exactly as before. Editing an existing draft is not re-gated either, matching `[wpuf_edit]`, which the renderer never gates.

**Helper choice:** `wpuf_is_option_on()` for the submission verdict rather than the literal `'yes' !==` comparison. Both producers — `Form::is_submission_open()` and the only `wpuf_can_post` filter (`Subscription::force_pack_permission`) — return strictly `yes`/`no`. `wpuf_is_checkbox_or_toggle_on()` would also accept `'1'` and `'true'`, which nothing emits for this value and which a security gate should not widen to accept; that helper stays correct for real toggle settings.

## Verification

- `php -l` clean on **7.4 and 8.5** for all three files. PHPCS: 0 errors on the changed lines.
- Role matrix and draft matrix above both executed against a live install (results as tabled).
- No hook, filter, template path or public method signature changed. New meta `_wpuf_draft_pending` is internal and self-clearing.

**End-to-end, in a real browser (Playwright against a fresh wp-env, free + pro on this branch):**

| step | result |
|---|---|
| Logged-out visitor registers through the form | account created, "Registration successful" — **legit path unaffected** |
| Anonymous nonce lifted with a cookie-less `fetch('/registration/')` | returned a valid nonce — confirms the researcher's point that the nonce authenticates nothing |
| Same registration POST replayed from a **logged-in** victim session | **no account created — guard holds** (the account-creation half, and therefore the session swap, never happens) |
| Re-register an already-registered email (logged out) | no duplicate, user count unchanged |
| WPUF login | works; login page renders the logged-in state |

**Confirmed not in scope of the guard:** profile editing runs through `Frontend_Account::update_profile()`, a different handler that `process_registration()` never touches, so a logged-in user editing their profile is unaffected.

Test accounts created during verification were removed afterwards.

## Security self-review of this diff

Checked that the fixes do not introduce anything new:

| check | result |
|---|---|
| Can a user set/clear `_wpuf_draft_pending` to skip the gate? | **No.** Meta keys come from the form's field definitions (`prepare_meta_fields()` keys off `$value['name']`), which the admin controls — a submitter cannot choose the key. |
| Is the new meta exposed? | Underscore-prefixed, so protected — not REST-exposed, not shown in custom fields UI. Self-clearing once the submission completes. |
| Input handling | `absint( wp_unslash( $_POST['post_id'] ) )`. |
| Guard ordering in `draft_post()` | `check_ajax_referer()` → logged-in check → **then** the gate, so it cannot be used as an unauthenticated probe. |
| Filter safety | `wpuf_registration_forbidden_capabilities` is cast and empty-checked, so a third-party filter returning a non-array cannot fatal, and returning empty cannot silently disable the check. |
| Fail-closed | Unknown/non-existent role → `default_role`; empty or `no` verdict → blocked. |
| Output / SQL | No new output and no SQL; the gate message goes back through the same `send_error()` path the existing gate already used. |
| Contracts | No hook, filter, template path or public method signature changed. |

**One residual, unchanged from before this PR and worth stating plainly:** a user re-submitting a post they already own to a *different* form is still treated as an edit and is not gated. That requires owning the post, creates no content they could not already create, and behaves exactly as 4.3.11 did — but if we want to close it too, the fix is to refuse a `post_id` whose recorded `wpuf_form_id` is not the form being submitted to. Happy to add that here if you would rather not leave it.

## Still to do (tracked in wpuf-pro#1688)

- Ship `dc3ac019` in the next release, with the `users_can_register` behaviour change called out in the changelog — sites that deliberately disable WP registration so WPUF is the only signup path will be affected, and it is worth deciding whether they get a supported filter.
- Manual regression pass before release: pack holder posts + saves drafts, expired-pack user edits an existing post, guest posting, Dokan/WC Vendors registration, custom-role registration, autologin on and off, multisite.
- Offer the researchers a pre-release build — they asked for one.

### DIFF
--- a/includes/Ajax/Frontend_Form_Ajax.php
+++ b/includes/Ajax/Frontend_Form_Ajax.php
@@ -183,18 +183,25 @@ class Frontend_Form_Ajax {
         // site-wide guest nonce can be replayed against a subscription-gated form to
         // create a post with no order and no pending-payment status.
         //
-        // A draft-backed submission still counts as new: it carries a post_id but sends
-        // wpuf_form_status "new" (same signal the update logic honours below), so key the
-        // gate on both so a replayed draft id cannot skip it.
-        $wpuf_form_status    = isset( $_POST['wpuf_form_status'] ) ? sanitize_text_field( wp_unslash( $_POST['wpuf_form_status'] ) ) : '';
-        $is_new_submission   = ! isset( $_POST['post_id'] ) || 'new' === $wpuf_form_status;
+        // Whether this is a new submission is decided from server state, never from a
+        // client-supplied field. Keying it on the request shape let a caller carry a
+        // post_id and omit wpuf_form_status to have the request treated as an edit, which
+        // skipped the gate entirely while the update branch still published the post.
+        //
+        // A draft that has not been submitted yet still counts as new: draft_post() flags
+        // it with _wpuf_draft_pending and that flag is cleared once the submission goes
+        // through, so finishing a draft is gated exactly once. Posts without the flag are
+        // genuine edits (including everything created before this flag existed) and are
+        // not re-gated, matching [wpuf_edit], which the renderer never gates.
+        $posted_post_id    = isset( $_POST['post_id'] ) ? absint( wp_unslash( $_POST['post_id'] ) ) : 0;
+        $is_new_submission = ! $posted_post_id || (bool) get_post_meta( $posted_post_id, '_wpuf_draft_pending', true );
 
         if ( $is_new_submission ) {
             [ $user_can_post, $submission_info ] = $form->is_submission_open( $form, $this->form_settings );
             $user_can_post                       = apply_filters( 'wpuf_can_post', $user_can_post, $form_id, $this->form_settings );
             $submission_info                     = apply_filters( 'wpuf_addpost_notice', $submission_info, $form_id, $this->form_settings );
 
-            if ( 'yes' !== $user_can_post ) {
+            if ( ! wpuf_is_option_on( $user_can_post ) ) {
                 wpuf()->ajax->send_error(
                     ! empty( $submission_info )
                         ? $submission_info
@@ -407,6 +414,13 @@ class Frontend_Form_Ajax {
             $this->update_post_meta( $meta_vars, $post_id );
             // set the post form_id for later usage
             update_post_meta( $post_id, self::$config_id, $form_id );
+
+            // The submission went through, so the post is no longer an unsubmitted draft.
+            // Clearing the flag means later edits are treated as edits and are not gated
+            // again, which keeps someone whose pack has since expired able to edit their
+            // own content.
+            delete_post_meta( $post_id, '_wpuf_draft_pending' );
+
             // if user has a subscription pack
             $this->wpuf_user_subscription_pack( $this->form_settings, $post_id );
             // set the post form_id for later usage
--- a/includes/Frontend/Frontend_Form.php
+++ b/includes/Frontend/Frontend_Form.php
@@ -203,6 +203,31 @@ class Frontend_Form extends Frontend_Render_Form {
             wp_send_json_error( [ 'message' => __( 'You must be logged in to save drafts.', 'wp-user-frontend' ) ] );
         }
 
+        // Saving a new draft is part of the add-post flow, so it has to pass the same
+        // submission gate the renderer applies before it will show the form at all
+        // (Frontend_Form::add_post_shortcode()). Without this, a subscription-gated form
+        // could be driven through the draft endpoint to create a post that never passed
+        // the gate, which submit_post() would then finish.
+        //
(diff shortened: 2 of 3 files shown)


arifulhoque7:
Note on the failing **PHPCS inspection** check:

It is failing on **pre-existing** coding-standard debt in the two legacy files this PR touches, **not on anything this PR added**:

- `includes/Frontend/Registration.php` — 41 errors on the base `develop` commit (dating to `1b41649d`, June 2023)
- `includes/Ajax/Frontend_Form_Ajax.php` — 24 errors on base
- `includes/Frontend/Frontend_Form.php` — 0 errors on base and 0 after this PR

The workflow lints each changed file in full (from the PR files API), so touching a legacy file surfaces its existing debt. I mapped every PHPCS error to the diff: **0 fall on the lines this PR adds.**

This is the same situation as the last security PR on this file, #1951, which also showed `Run PHPCS inspection: fail` and was merged — PHPCS is advisory here (not a required check on `develop`). The `guard` check, which enforces that built assets stay in sync, passes.

Deliberately not running `phpcbf` across these files in this PR: it would add a large, behaviour-adjacent formatting diff (including removing "unreachable code after return" that may be intentional) to a security change that should stay small and reviewable. The 2023 debt is worth a separate cleanup PR.

PR #1972, opened 2026-09-07
fix: admin notice instead of a fatal when Composer deps are missing

arifulhoque7:
## Problem

A fresh clone that has not run `composer install` white-screens the whole site the moment WP User Frontend loads:

```
PHP Fatal error: Trait 'WeDevs\WpUtils\SingletonTrait' not found in wp-user-frontend/wpuf.php on line 40
```

`vendor/` is intentionally gitignored, so `wpuf.php` loads `vendor/autoload.php` only when it exists, but then **unconditionally** declares a class that `use`s `SingletonTrait` from `wedevs/wp-utils`. With no autoloader the trait is undefined and the class declaration is a hard fatal, before any admin notice can render.

## Root cause is the missing guard, not the lock

Verified in a clean-room clone: `composer install` from the committed `composer.lock` installs `wedevs/wp-utils` (dev-main) correctly and the plugin loads. The fatal only happens when someone activates the plugin **without** running `composer install`.

| clean clone | before this PR | after this PR |
|---|---|---|
| no `composer install` (no vendor/) | **white-screen fatal, line 40** | admin notice "run `composer install`", site stays up |
| after `composer install` | loads | loads (unchanged) |

## Fix

Guard the bootstrap: if `SingletonTrait` is not available, register an `admin_notices` message telling the user to run `composer install`, and `return` early before the class that needs it is declared. When `vendor/` is present nothing changes.

Additive, one file, no behavior change for a correctly-installed site. `php -l` clean.

https://claude.ai/code/session_019KGP9sp935QBeu32pGt5WY


arifulhoque7:
Closes weDevsOfficial/wpuf-pro#1694

Issue #1945, opened 2026-08-20
Post Expiration is applied even when disabled


006rodaal:
We are experiencing an issue with the Post Expiration functionality in WP User Frontend Pro.

Versions
WP User Frontend: 4.3.10
WP User Frontend Pro: 4.2.16

Post Expiration is disabled in the settings of these forms.

For example, the serialized wpuf_form_settings for a form contains:
expiration_settings:
enable_post_expiration = off
expiration_time_value = 7
expiration_time_type = day
expired_post_status = draft

So the saved configuration correctly indicates that Post Expiration is disabled. However, newly created posts are still receiving the following post meta:
wpuf-post_expiration_date
wpuf-expired_post_status

For example, a new post created on August 20, 2026 received:
_wpuf_form_id = 27
wpuf-post_expiration_date = 2026-08-27
wpuf-expired_post_status = draft

This is exactly 7 days after the post was created, matching the old expiration settings stored in the form configuration. As a result, published posts can automatically be changed to Draft after 7 days, despite Enable Post Expiration being disabled.

We have confirmed that this is not limited to old posts that were created while Post Expiration was enabled. A completely new post, created more than a week after Post Expiration had been disabled, immediately received the expiration metadata.

Expected behavior
When:
enable_post_expiration = off
new posts should not receive an expiration date and should not later be changed to Draft by the Post Expiration functionality.

Actual behavior
WPUF appears to still use:
expiration_time_value = 7
expiration_time_type = day
expired_post_status = draft

even though:
enable_post_expiration = off

Could you please confirm whether this is a known issue and whether there is a fix available?

We would also like to know whether the version combination of WP User Frontend 4.3.10 and WP User Frontend Pro 4.2.16 could be related to this behavior.

anik-fahmid:
@arifulhoque7 vai, please check

arifulhoque7:
Confirmed and fixed. Root cause: `save_expiration_meta()` guarded on `empty( enable_post_expiration )`, but the form stores the toggle as the string `'off'` when disabled and `empty('off') === false` — so the guard passed and every new post got `wpuf-post_expiration_date` / `wpuf-expired_post_status` from the stale time value, flipping published posts to Draft.

Fix (wpuf-pro): guard on the real on/off value via `wpuf_is_checkbox_or_toggle_on()`.

Verified on a live site by firing the real submit hook:
```
before: enable=off => META WRITTEN
[...]

PR #1701, opened 2025-09-12
Feature/dynamic field icons

arifulhoque7:
### PR: Add Icon Support for Form Fields in WPUF Form Builder  Close [1048](https://github.com/weDevsOfficial/wpuf-pro/issues/1048) Related [PR](https://github.com/weDevsOfficial/wpuf-pro/pull/1122) in pro

**Description**
This PR enhances the WP User Frontend (WPUF) form builder by introducing an **icon option** for form fields. Admins can now visually improve their forms by placing icons next to field labels or inside input fields (prefix/suffix). The feature is designed to boost usability, accessibility, and overall design consistency.

**Key Features**

* Added **"Field Icon" option** in form builder settings for all fields.
* Integration with **Font Awesome 7 Free** (supporting **2,806 icons**).
* Placement options:

  * No icon (default)
  * Left of label
  * Right of label
  * Inside input (prefix)
  * Inside input (suffix)
* Icons render seamlessly on the frontend across **all WPUF form types** (post forms, registration forms, etc.).
* Backward compatible (no changes for existing forms without icons).

**Benefits**

* Enhances form aesthetics and modernizes UI.
* Improves user experience by giving visual cues for field purposes (e.g., 📧 email, 📞 phone, 🔑 password).
* Consistent with modern UI/UX patterns.

**Considerations**

* Optional per field, so existing forms remain unaffected.
* Icons load from Font Awesome CDN; minimal performance impact.
* Works with both **WPUF Free and Pro**.
* No additional color customization included (can be styled with theme/custom CSS).

**Acceptance Criteria**

* [x] Icons can be added or removed per field in the form builder.
* [x] Icons appear correctly on frontend forms (label & input placements).
* [x] No conflicts with themes or WPUF custom CSS.
* [x] Tested across WPUF Free and Pro modules.

## Summary by CodeRabbit

- New Features
  - Added icon selector with search and grid to choose field icons.
  - Support showing icons on form field labels, positioned left or right.
  - Default icons now auto-apply when enabling icons; new fields get icon defaults.

- Bug Fixes
  - Removed duplicate name attributes on radio options for consistent grouping.

- Style
  - New icon styling for labels and spacing/utilities.

- Chores
  - Updated icon library and made icon sets configurable for the form builder.

PR #1971, opened 2026-09-07
fix(security): Patchstack 4.3.11 — file-deletion IDOR, guest pay bypass, PayPal lifecycle abuse

arifulhoque7:
Fixes the three Patchstack reports against 4.3.11. I reproduced all three **live** first (PoC harnesses driving the real code paths on a running install), fixed them, and re-ran each to confirm it is closed while the legitimate flow still works.

| # | Report | CVSS | Root cause | Reproduced | After fix |
|---|--------|------|-----------|-----------|-----------|
| 1 | wpuf-pro#1691 Subscriber+ arbitrary file deletion | 6.5 | attachment reparent + over-limit delete, both without ownership checks | admin attachment destroyed by a Subscriber | attachment survives; foreign reparent skipped |
| 2 | wpuf-pro#1692 unauth guest pay-per-post publish bypass | 5.3 | `p_id`/`f_id` never bound; author-0 via nonce-param confusion | pending unpaid post published | stays pending, redirects to payment |
| 3 | wpuf-pro#1693 recurring pack on unapproved/unpaid PayPal CREATED | 6.5 | grant on `APPROVAL_PENDING`; wrong quota key | limited pack granted **unlimited**, no payment | no grant until `ACTIVE`; limited quota preserved |

## 1. Arbitrary file deletion (wpuf-pro#1691)

Two chained IDORs. The `post_content` `` handler reparented **any** same-site attachment to the attacker's post with no ownership check; the file field's over-limit loop then passed that now-"associable" id straight to `wp_delete_attachment()`.

- `submit_post()` — reparent an embedded image only when `is_attachment_associable()` (already this post's, or unattached **and** owned by the requester), the same guard the featured-image path already uses.
- `FieldableTrait` file loop — delete an over-limit attachment only when the requester actually owns it or can delete it.
- `submit_post()` — reject a `form_id` that is not a real `wpuf_forms` post, which also closes the reported "action-wide nonce → arbitrary published post" case.

## 2. Guest pay-per-post publish bypass (wpuf-pro#1692)

`p_id` and `f_id` are encrypted independently and never bound, so replaying the post-id token as `f_id` ran the paywall against a non-form object with empty settings and published the post for free while payment stayed pending. A nonce-parameter confusion produced the author-0 post the verifier looks for.

- `publish_guest_post()` — derive the charging decision from the post's own stored `_wpuf_form_id`, reject an `f_id` that does not match it, require a real `wpuf_forms` id. The paywall is always evaluated against the real form.
[...]

PR #1702, opened 2025-09-12
Feat/wpuf ai form builder design

arifulhoque7:
Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1083) , Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1085) , Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1087) ,Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1088) , Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1113), Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1114) , Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1115) , Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1116), Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1089) ,Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1144) , Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1175)

Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1154), Close [issue](https://github.com/weDevsOfficial/wpuf-pro/issues/1157)
### 📋 Overview

This PR introduces a groundbreaking **AI-powered form builder** for **WP User Frontend**, allowing users to create complex WordPress forms through natural language descriptions. It reimagines the traditional form creation workflow by integrating a conversational, intelligent interface that adapts to user intent and provides real-time refinement options.

The feature is built with a mix of **AI orchestration**, **Vue.js components**, **modern styling**, and **secure REST APIs**, transforming WP User Frontend into a next-generation, AI-assisted platform.

---

#### 🧠 AI Infrastructure (+8,500 lines)

* **Core AI System** – Multiple provider support with centralized orchestration.
* **Predefined Provider** – 1,189 lines of intelligent form pattern recognition.
* **Form Generator** – 577 lines of context-aware generation logic.
* **REST API Controller** – 1,004 lines of secure API endpoints.
* **AI Manager** – Unified service handling for AI interactions.

#### 🖥️ Vue.js Components (+2,825 lines)

* **AIFormBuilder.vue** – Main orchestration with 3-stage workflow.
* **FormInputStage.vue** – Natural language form input interface.
* **FormProcessingStage.vue** – Real-time form generation progress feedback.
[...]

PR #1980, opened 2026-09-18
fix: stop post expiration from drafting posts it should leave alone (#1701)

arifulhoque7:
Closes weDevsOfficial/wpuf-pro#1701

Free-side half of the post-expiration work found while investigating weDevsOfficial/wpuf-pro#1700. The pro half is weDevsOfficial/wpuf-pro#1702. Either can merge first; neither depends on the other to build.

## Problem

Three defects here make WPUF write an expiry date the daily `wpuf_remove_expired_post_hook` cron then acts on, unpublishing posts the site owner never asked to expire.

### 1. The stored pack duration is unparsable

`Subscription::save_form_meta()` built the duration as:

```php
$expiration_time = <value> . ' subscription.php' . <type>;
```

A stray `' subscription.php'` literal had replaced the separating space, so the pack editor stored `"7 subscription.phpday"`. `strtotime()` returns `false` for that and `gmdate( 'Y-m-d', false )` renders `1970-01-01`, so posts submitted under that pack were already past their expiry date and the next cron run drafted them. The sibling path in `Admin_Subscription.php` uses the correct `' '` separator, which is what the runtime expects.

### 2. A pack with expiration switched off still expired posts

`Frontend_Form_Ajax::wpuf_user_subscription_pack()` gated on `isset()` and never looked at the value, so a pack storing `'off'` still wrote the metadata. Same class of bug as #1945, which was only fixed on the pro side.

### 3. No guard on the `strtotime()` result

The stored duration went straight into `strtotime()`. Empty or unparsable input gave `false`, producing the `1970-01-01` date above.

## Fix

- Restore the single-space separator so the pack stores a duration `strtotime()` can read.
- Gate the pack branch on the actual value via `wpuf_validate_boolean()`. That is the most permissive truthiness test that still rejects `'off'`, `''` and `'0'`, chosen over `wpuf_is_option_on()` so a pack storing something like `'1'` keeps working rather than silently losing expiration.
- Only write the metadata once `strtotime()` has returned a real timestamp.

## Testing

Clean site, only WPUF 4.3.11 + Pro 4.2.17 active, no other plugins. Posts submitted by filling the real frontend form in a browser, then the registered cron run with `wp cron event run wpuf_remove_expired_post_hook`.

Before:

| pack | expiry written | after cron |
|---|---|---|
| `off`, time `''` | `1970-01-01` | published to **draft** |

After, 21 of 21 assertions pass:

| case | expiry written | after cron |
|---|---|---|
| pack `off`, time `7 day` | none | stays published |
| pack `''`, time `7 day` | none | stays published |
| pack `on`, time `''` | none | stays published |
| pack `on`, corrupt time | none | stays published |
| pack `on`, time `7 day` | +7 days | stays published |
| pack `yes`, time `1 month` | +1 month | stays published |
| pack editor saves `7` + `day` | stores `"7 day"`, parsable | n/a |
| past-dated post | n/a | unpublished, unchanged |
| future-dated post | n/a | stays published |

Submission, redirect and notification mail were smoke-tested through the browser after the formatting pass below.

## Notes

- No hook, filter, method signature or meta key changed.
- LF line endings preserved; no tab indentation introduced.
- The diff has two distinct parts. The functional change is in `wpuf_user_subscription_pack()` and the one separator line in `Subscription.php`. The rest is a PHPCS cleanup of violations that already existed in these two files, needed because CI lints whole changed files rather than the diff: the empty `if ( empty( $to ) ) {} else {}` pairs became `if ( ! empty( $to ) )`, and `else { if }` pairs became `elseif`. Both files now report 0 PHPCS errors, down from 30.
- **This does not clean metadata already written.** See the note on weDevsOfficial/wpuf-pro#1702 for the support cleanup query.


dev-shahed:
## ⚠️ QA — merge-order hazard with #1971 (blocking)

This PR and #1971 both rewrite `includes/Ajax/Frontend_Form_Ajax.php` from the same base blob in a PHPCS sweep, and neither carries the other's fix:

| ref | blob | carries |
|---|---|---|
| `develop` | `cf951cae8f` | neither |
| **#1971** | `43c7a9861b` | `Invalid form.` guard, `is_attachment_associable()`, `_ajax_nonce` fix — **no** expiration gate |
| **this PR** | `bf26428b22` | `wpuf_validate_boolean()` gate + `strtotime()` guard — **none** of #1971's guards |

Test-merge confirms it: the other five free PRs merge clean, #1971 then conflicts here and in `FieldableTrait.php`. Whichever lands second silently drops the other's change — no conflict marker, no failing check.

**Ask:** sequence with #1971, rebase the second onto `develop`, then assert on `develop` that a pack with `_enable_post_expiration = 'off'` writes no expiry **and** `submit_post()` still rejects a non-`wpuf_forms` `form_id`.

Fix itself verified end to end on a live install (pack `off`/empty/corrupt duration → no expiry, post stays published after the real cron; valid packs unchanged). Two notes for the record: the Pro half (#1702) is red on CI because `includes/Post_Expiration.php` is CRLF on `develop` already, and neither PR cleans metadata already written — affected sites keep `1970-01-01` and keep getting drafted.

arifulhoque7:
Update: the overlap is gone. `293ddbbd` on #1971 (and `c1d33250` on #1963) make the branches use identical text on every line they had each reformatted, so nothing conflicts any more.

Re-measured on the current heads: all ten open PRs I own merge onto `develop` clean in ascending order, clean in descending order, and clean across all **90 ordered pairs**. Merge order no longer matters.

Assertions on the fully stacked tree, run live: legacy repeat stores the sanitized string; over-limit delete keeps the ownership guard; `submit_post()` rejects a non-`wpuf_forms` `form_id`; a pack with expiration `off` and an unparsable duration both write no expiry meta, a valid duration still does; `php -l` clean.

Full detail in https://github.com/weDevsOfficial/wp-user-frontend/pull/1971#issuecomment-5756569928

